PR #14, opened 2019-05-29
Refactor code and use rsuite instead of antd

rehandalal:
r?


mythmon (changes requested):
This looks a lot better, thanks for the updates. I have a couple issues:

* The recipe list page grows wider than the screen when i expand recipes with long lines in them. Can we put a max width on the contents of the expanded recipes so that it doesn't make a horizontal scroll bar?
* The filter page stacking vertically seems like a regression to me, since as the tree expands the text box moves around. I don't think that was on purpose though.
* The new icon looks great (I like how it is customizable in the React component). I think the 16px size is a bit of a regression though. I think we could leave it as is and add yet another icon for 16px later.

mythmon (approved):
This is looking great. Thanks for the changes :tada:

PR #19, opened 2019-10-18
More recipe running

rehandalal:
yeesh, finally!

r?


mythmon:
I poked at it for a second, and I was able to do both of the new things. I didn't look at the code at all. I'd rather merge this now and iterate than hold this up in review.

:see_no_evil:

PR #25, opened 2019-11-22
Recipe Timeline

mythmon:
I added a new page, and in doing so made a base component for new pages. That new page is a timeline of events for a recipe. I've been wanting to make this kind of thing to automate the analysis I did in mozilla/normandy#2040. The devtools seemed like a good place to do it.

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,6 @@ module.exports = {
   ],
   rules: {
     "mozilla/no-define-cc-etc": "off", // seems broken outside of m-c
-    "generator-star-spacing": "off",
+    "generator-star-spacing": "off", // this is handled by Prettier
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,5 +9,6 @@ module.exports = {
   ],
   rules: {
     "mozilla/no-define-cc-etc": "off", // seems broken outside of m-c
+    "generator-star-spacing": "off",
   },
 };
--- a/extension/content/components/pages/AddonStudiesPage.js
+++ b/extension/content/components/pages/AddonStudiesPage.js
@@ -2,15 +2,15 @@ import React from "react";
 import { Loader } from "rsuite";
 
 import AddonStudy from "devtools/components/studies/AddonStudy";
-import BasePage from "./BasePage";
+import BasePage from "devtools/components/pages/BasePage";
 
 const normandy = browser.experiments.normandy;
 
-export default class PrefStudiesPage extends BasePage {
+export default class PrefStudiesPage extends React.PureComponent {
   constructor(props) {
     super(props);
     this.state = {
-      studies: [],
+      studies: null,
       loading: true,
     };
   }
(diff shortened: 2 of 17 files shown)


extension/content/components/pages/AddonStudiesPage.js
```diff
@@ -2,10 +2,11 @@ import React from "react";
 import { Loader } from "rsuite";
 
 import AddonStudy from "devtools/components/studies/AddonStudy";
+import BasePage from "./BasePage";
```
rehandalal: nit: use absolute import path

extension/experiments/normandy/api.js
```diff
@@ -87,7 +87,12 @@ var normandy = class extends ExtensionAPI {
           },
 
           async checkRecipeFilter(recipe) {
-            return RecipeRunner.checkFilter(recipe);
+            if (RecipeRunner.checkFilter) {
+              return RecipeRunner.checkFilter(recipe);
+            } else if (RecipeRunner.shouldRunRecipe) {
+              return RecipeRunner.shouldRunRecipe(recipe);
+            }
+            throw new Error("Incompatible version of Normandy");
```
rehandalal, line 95: 👏🏽 one thing i'd love to look at as we grow this addon is regression testing against nightly to catch stuff like this early.

.eslintrc.js
```diff
@@ -9,5 +9,6 @@ module.exports = {
   ],
   rules: {
     "mozilla/no-define-cc-etc": "off", // seems broken outside of m-c
+    "generator-star-spacing": "off",
```
rehandalal: can we pick something for a consistent spacing of the star rather than turning this off?

rehandalal: i think im slightly partial to `both`

extension/content/components/pages/AddonStudiesPage.js
```diff
@@ -2,10 +2,11 @@ import React from "react";
...
 
 const normandy = browser.experiments.normandy;
 
-export default class PrefStudiesPage extends React.PureComponent {
+export default class PrefStudiesPage extends BasePage {
```
rehandalal: the official take on inheritance: https://reactjs.org/docs/composition-vs-inheritance.html

extension/content/components/pages/RecipesPage.js
```diff
@@ -1,170 +1,66 @@
...
-  Pagination,
-  SelectPicker,
-} from "rsuite";
+import { Button, Icon, Loader, Modal, Nav, Pagination } from "rsuite";
+import { Map } from "immutable";
 
 import { ENVIRONMENTS } from "devtools/config";
 import RecipeListing from "devtools/components/recipes/RecipeListing";
-import api from "devtools/utils/api";
+import BaseApiPage from "./BaseApiPage";
```
rehandalal: nit: no relative import paths

extension/content/components/recipes/RecipeTimeline.js
```diff
@@ -0,0 +1,126 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Timeline, Icon } from "rsuite";
+import moment from "moment";
+
+import "./RecipeTimeline.less";
```
rehandalal: nit: use absolute

rehandalal (changes requested):

.eslintrc.js
mythmon: Prettier (from `extends: [..., "plugin:prettier/recommended"]`) handles generator star spacing. This is just getting ESLint to stop fighting with it. This ESLint rule should have been disabled by the ESLint/Prettier integration, I'm not sure why it wasn't.

extension/content/components/pages/AddonStudiesPage.js
mythmon: I'll switch it over.

mythmon (commented):
Ack, sorry about the relative imports. VSCode has some very handy automatic import logic, but it's sort of invisible and it makes relative imports. I'll clean those up.

mythmon:
This is sort of ready to go. If you have time to review it, go for it. I think I might split this into two PRs though, one for the basepage stuff and one for the timeline page. I think that will make things simpler overall.

mythmon:
I'm actually going to abandon this. I don't think it's worth fixing the merge conflicts and cleaning this up. I think the idea of a common base page is a good idea though. That's probably something for the editing refactor though. I think the idea of a base page is something we should use in the future.

PR #27, opened 2019-11-29
Optimize

rehandalal:
Some tweaks to take advantage of tree-shaking and bring the bundle size down from 6M to 3.3M:
- Upgrade `rsuite` and use the new `babel-preset-rsuite` to facilitate tree-shaking.
- Had to ditch `react-highlight` since it breaks tree-shaking for `highlight.js` and write a custom component to wrap `highlight.js`.
- The newest version of rsuite injects a global in JS and I couldn't figure out how to get webpack to be less obnoxious about how it deals with that so I just updated the CSP.

Also upgrade all the dependencies and pin their versions.

r?


mythmon (approved):
r+ but the tests are grumpy

rehandalal:
Also added `bors.toml` so we can have bors support going forward.

rehandalal:
oh actually just noticed the r+

PR #29, opened 2019-12-05
Add protocol handler and pretty address bar for pretty, shareable urls

mythmon:
This is admittedly pretty silly right now, but I think it will be helpful in the long term. Having links we can use to communicate about recipes when this replaces Delivery Console will be really useful.

The extra address bar is mostly for show, and so that we have an easy place to copy from. These URLs also work in the real address bar.

With this version of the add-on installed, you can have things like links to ext+normandy://recipes that actually work in web content. Well, they would except Github won't let me embed that kind of link. It totally works [on JSBin](https://jsbin.com/qicobacewe/1/edit?html,output) though.

image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -283,10 +283,17 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
-"@babel/runtime@^7.0.0", "@babel/runtime@^7.1.2", "@babel/runtime@^7.4.0":
+"@babel/runtime@^7.0.0":
   version "7.4.5"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.4.5.tgz#582bb531f5f9dc67d2fcb682979894f75e253f12"
   integrity sha512-TuI4qpWZP6lGOGIuGWtp9sPluqYICmbk8T/1vpSysqJxRPkudh/ofFWyqdcMsDf2s7KvDL4/YHgKyvcS3g9CJQ==
+  dependencies:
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -6423,32 +6425,32 @@ react-highlight@^0.12.0:
...
 react-lifecycles-compat@^3.0.4:
   version "3.0.4"
   resolved "https://registry.yarnpkg.com/react-lifecycles-compat/-/react-lifecycles-compat-3.0.4.tgz#4f1a273afdfc8f3488a8c516bfda78f872352362"
   integrity sha512-fBASbA6LnOU9dOU2eW7aQ8xmYBSXUIWr+UmF9b1efZBazGNO+rcXT/icdKnYm2pTwcRylVUYwW7H1PHfLekVzA==
 
 react-router-dom@^5.0.1:
-  version "5.0.1"
-  resolved "https://registry.yarnpkg.com/react-router-dom/-/react-router-dom-5.0.1.tgz#ee66f4a5d18b6089c361958e443489d6bab714be"
-  integrity sha512-zaVHSy7NN0G91/Bz9GD4owex5+eop+KvgbxXsP/O+iW1/Ln+BrJ8QiIR5a6xNPtrdTvLkxqlDClx13QO1uB8CA==
+  version "5.1.2"
```
mythmon, line 6438: This is the driving change here. `react-router-dom@5.1.0` introduced the hooks I used for the address bar.

rehandalal (approved):

PR #31, opened 2020-01-23
Configure web-ext to set needed prefs

mythmon:
With recent changes to Firefox, we now need to set `extensions.experiments.enabled=true` in order to use web-ext to test the add-on. This configuration file automates that, as well as automatically opening the browser console, which can be quite useful.


rehandalal (approved):
bors r+

PR #32, opened 2020-01-23
 Load filter context dynamically instead of with a static list 

mythmon:
This is more future proof, since it will automatically pick up normal attributes that get added to the filter context. The class walking code is derived from how Normandy Client [calculates the available properties of the context](https://searchfox.org/mozilla-central/rev/f98dad153b59a985efd4505912588d4651033395/toolkit/components/normandy/lib/RecipeRunner.jsm#422-446) for capability calculation.


rehandalal (approved):
r+

PR #34, opened 2020-01-29
try using level 3 docker images

escapewindow:


rehandalal (approved):
r+

escapewindow:
@rail because we don't have anything creating the level 1 docker image, that index expired. I think it's ok to reuse the level 3 docker image, just as an environment to build and test in.

PR #37, opened 2020-02-14
recipe view button/modal

tiftran:


rehandalal (approved):
nice work! thanks for this

bors r+

extension/content/components/pages/RecipesPage.js
```diff
@@ -1,6 +1,7 @@
 import autobind from "autobind-decorator";
 import React from "react";
 import { Controlled as CodeMirror } from "react-codemirror2";
+import Highlight from "react-highlight";
```
rehandalal: oops! this should be:
```js
import Highlight from "devtools/components/common/Highlight";
```

rehandalal (changes requested):
needs an import path updated

mythmon (dismissed):
Looks like you added a new dependency, `react-highlight`, but didn't add it to `package.json`? If you're using Yarn, you can say `yarn add react-highlight`.

rehandalal:
> Looks like you added a new dependency, react-highlight, but didn't add it to package.json? If you're using Yarn, you can say yarn add react-highlight.

`react-highlight` was a dependency but was removed in #27 since it was breaking tree-shaking

extension/content/components/pages/RecipesPage.js
```diff
@@ -197,6 +198,7 @@ class RecipesPage extends React.PureComponent {
       this.setState({ runningArbitrary: false });
     }
   }
+  /* eslint-enable no-useless-catch */
```
rehandalal: this should actually go up above on line 200 above the last `}`.

tiftran: will do

rehandalal (approved):
bors r+

PR #51, opened 2020-03-24
Implement authentication using Auth0

### LINKED ISSUE #40: Auth0 authentication implementation
We need to find some way to authenticate with Auth0. A couple useful ideas:
- Use the popup workflow
- Use the `ext+normandy` protocol for the callback URL

### LINKED ISSUE #42: Handle Auth0 for multiple environments
As we will be connecting to multiple environments (prod, stage, dev) we will need ways to authenticate against each and keep track of which ones we are/aren't authenticated against.

In an ideal world we'd have a solution that works across environments.

rehandalal:
Fixes #40 
Fixes #42 

I think there's still some improvements needed on this:
- [x] ~I think creating a couple new contexts for `Environment` and `Auth` would probably be good.~ Filed a related issue #69 
- [x] ~Tokens are not refreshed.~ Filed a related issue #68
- [x] This went through a bunch of iterations and so there's probably some extra things we might not need. A once over to clean up would be a good idea.

Some extra notes:
- ~We will need the Auth0 config for all the environments to be updated to include `https://08e23a90be40fa842a9f18ac049d45a7c3c8e7ec.extensions.allizom.org/` in the Allowed Callback URLs. Until that is done you will need your own Auth0 tenant set up to test this PR.~ EDIT: Auth0 config was updated and all the environments should work now.
- You can create a `.env` file in the root to configure the local environment if you are using anything other than standard settings. I'll file a new issue for adding UI to edit environment configurations through NDT itself.

r?


extension/content/utils/environmentStore.js
```diff
@@ -0,0 +1,90 @@
...
+    this.writeableUrl = config.writeableUrl;
+    this.oidcClientId = config.oidcClientId;
+    this.oidcDomain = config.oidcDomain;
+    this.authSession = new AuthSession(key, config);
+    this.authSession.registerHandler("login", this.handleLogin);
+    this.authSession.registerHandler("logout", this.handleLogout);
+  }
+
+  handleLogin(authResult) {
+    this.setAuthToken();
```
mythmon: It looks like this method doesn't exist?

extension/content/components/common/AppHeader.js
```diff
@@ -0,0 +1,194 @@
...
+  const [showEnvironmentModal, setShowEnvironmentModal] = React.useState(false);
+
+  const environments = environmentStore.getAll();
+  const envOptions = Object.keys(environments).reduce((reduced, value) => {
+    reduced.push({
+      label: value.charAt(0).toUpperCase() + value.slice(1),
+      value,
+    });
+    return reduced;
+  }, []);
```
mythmon: Why not `Object.keys(evironments).map(...)`?

extension/content/components/common/AppHeader.js
```diff
@@ -0,0 +1,194 @@
...
+      onClick={() => {
+        environment.authSession.login();
+      }}
+    >
+      Log In
+    </IconButton>
+  );
+}
+
+export default function AppHeader() {
```
mythmon: It's a minor thing, but I think we should put the "main" component of a file that consumes the rest at the top of the file. It helps put things into context, and usually that's what I'd be looking for when I open this file.

extension/content/components/common/AppHeader.js
```diff
@@ -0,0 +1,194 @@
...
+
+  return (
+    <InputGroup>
+      <InputGroup.Addon>
+        <Icon icon="link" />
+      </InputGroup.Addon>
+      <Input
+        type="text"
+        value={address}
+        onChange={ev => setAddress(ev.target.value)}
```
mythmon: I can't type in the address bar. I get `TypeError: can't access property "value", ev.target is undefined` on this line.

extension/content/components/pages/RecipesPage.js
```diff
@@ -1,45 +1,47 @@
...
+import environmentStore from "devtools/utils/environmentStore";
 import { convertToV1Recipe } from "devtools/utils/recipes";
+import { globalStateContext } from "devtools/contexts/globalState";
 
 const normandy = browser.experiments.normandy;
 
 @autobind
 class RecipesPage extends React.PureComponent {
+  static propTypes = {
+    environment: PropTypes.string,
```
mythmon: I get 

> Failed prop type: Invalid prop `environment` of type `object` supplied to `RecipesPage`, expected `string`.

on this line. Is this the right prop type?

extension/content/contexts/globalState.js
```diff
@@ -0,0 +1,45 @@
...
+        ...state,
+        selectedEnvironment: environmentStore.get(action.environment),
+      };
+
+    default:
+      return state;
+  }
+};
+
+export const GlobalStateProvider = ({ children }) => {
```
mythmon: Can we avoid the `const name = () => {}` style and use `function name() {}` instead for top-level functions?

extension/content/utils/authSession.js
```diff
@@ -0,0 +1,263 @@
...
+      localStorage.getItem(`environment.${this.key}.auth.result`),
+    );
+  }
+
+  get accessToken() {
+    return this.authResult && this.authResult.accessToken;
+  }
+
+  get profile() {
+    return this.authResult && this.authResult.idTokenPayload;
```
mythmon: This pattern means that we're calling `localStorage.getItem` and `JSON.parse` are called twice. Can we make this only call those two once?

extension/content/utils/authSession.js
```diff
@@ -0,0 +1,263 @@
...
+    this.handlers[event] = this.handlers[event].filter(h => h !== handler);
+  }
+
+  /**
+   * Generate a random string to use as a nonce.
+   *
+   * @param {int} length  The length of the generated string.
+   * @returns {string}
+   */
+  generateNonce(length) {
```
mythmon, line 87: I have a bunch of questions about this. Mostly, where did it come from? Should we be "improving" it, or should we leave it as is because it came from someone like Duo? If so, we should probably leave a comment to that effect.

mythmon (changes requested):
There are a couple places that seem like bugs, and I have a bunch of other questions as well. If you could answer the questions first, I could take over on the bug fixing.

@tiftran I think it is still valuable for you to review this as well.

extension/content/utils/authSession.js
rehandalal: I'll add a comment but it came from Auth0's docs: https://auth0.com/docs/api-auth/tutorials/nonce#generate-a-cryptographically-random-nonce

They also mention using [nanoid](https://github.com/ai/nanoid) so that's another option.

mythmon (approved):

mythmon:
I'd like to hear from @tiftran before merging this.

tiftran:
eek! I missed this yesterday...

Was wondering if [this](https://en.wikipedia.org/wiki/Cryptographic_nonce#/media/File:Nonce-cnonce-uml.svg) describes what's happening with auth0 correctly? and is it a token that comes back from auth0 and then is set to local storage?

Also got to read up about context which was cool 🙂

mythmon:
Hm. I don't think that image describes what's happening here. The nonce is generated by the client, and is used as a message ID so that we know the response from Auth0 is the right one. I think this is probably closer:

mermaid

Disclaimer, OAuth is complicated and there is a lot more to it than this. I'm not really qualified to explain it. I think this is enough for this PR though.

tiftran (commented):
Looks good to me?
~~When I checked it out and ran it locally and I noticed some minor ux thing? when you log in and out, it's not necessarily clear if you successfully done so until you manually refresh the page?~~ which you fixed in #70...

mythmon:
bors r= @mythmon, @tiftran

PR #70, opened 2020-03-29
Implement context for environment and auth

### LINKED ISSUE #69: Convert global state context to individual contexts for environment and auth
Instead of using external stores and a global state to pull from them we should maintain all the data we care about in individual contexts.

Depends on #51 landing.

rehandalal:
Depends on #51.

Fixes #69.

r?


extension/content/utils/api.js
```diff
@@ -1,30 +1,111 @@
...
+      headers.append("Content-Type", "application/json");
+    }
+
+    for (let headerName in extraHeaders) {
+      headers.append(headerName, extraHeaders[headerName]);
+    }
+
+    const settings = {
+      headers,
+      credentials: "same-origin",
```
mythmon, line 28: 1. I think `same-origin` is the default now
2. "Credentials" here refers to Cookies, which we aren't using
3. We won't be on the same origin anyways, since we are a web extension.

I think we should remove this.

mythmon (approved):
I think I understand this. It turns environments and auth into a single global provider, and then exposes a bunch of React Hooks to interact with it, and wraps it around the top level app. Then it wraps the recipe page in a functional component so it can translate the hooks for a class-based component. I haven't run it myself, but lets go with it to keep things unblocked.

bors merge

PR #71, opened 2020-03-29
Automatically select light or dark theme based on system preference

### LINKED ISSUE #53: Automatic dark and light theme support
I have switched to the dark theme from the new version of rsuite but ideally we should have the theme picked by the users settings

rehandalal:
Depends on #51.

Fixes #53.

r?


mythmon (approved):
I don't know how to change my preferred color scheme, but I see the add-on in light mode now, so I assume it works.

bors merge

PR #72, opened 2020-03-29
Recipe editor

### LINKED ISSUE #54: Basic recipe editing page
We need to have a basic recipe editing page that has all the necessary fields to get a barebones recipe together. It should have fields for:
- Name
- Experimenter Slug
- Extra Filter Expression -> generic text area 
- Action
- Action Arguments (generic textarea)

We will build on this in subsequent issues

tiftran:
Fixes #54


extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,188 @@
...
+      arguments: arguments_,
+    } = recipe.latest_revision;
+    const action_id = recipe.latest_revision.action.id;
+    setData({
+      name,
+      experimenter_slug,
+      filter_expression,
+      action_id,
+      arguments: JSON.stringify(arguments_, null, 1),
+    });
```
tiftran: this seems really messy.... but not sure how to make this more clean...

extension/content/utils/request.js
```diff
@@ -0,0 +1,72 @@
+export async function request(url, options = {}, extraHeaders = {}) {
```
tiftran: and this was taken from experimenter... 😐

rehandalal: I ended up writing a version of this in #70

extension/content/components/recipes/RecipeEditor.js
rehandalal: you should probably just store the unedited revision as a whole here. it doesn't matter if you have extra keys that aren't being used by the form. you should stringify the arguments in the arguments field.

extension/content/components/pages/RecipesPage.js
```diff
@@ -266,22 +306,21 @@ class RecipesPage extends React.PureComponent {
...
-              pages={Math.ceil(count / 25)}
-              onSelect={this.handlePageChange}
-              size="lg"
-              prev
-              next
-              first
-              last
-              ellipsis
-              boundaryLinks
+          <Switch>
```
rehandalal: I think that we should do all routing at the `App` level as we currently do.

rehandalal (changes requested):
nice work!

a couple follow up tasks:
- write some tests
- merge in the latest master which has the API changes I made

extension/content/utils/api.js
```diff
@@ -97,6 +97,13 @@ export default class NormandyAPI {
       version,
     });
   }
+
+  async fetchActions({ version = 3 }) {
```
rehandalal: there is no other version of action endpoint so we don't need the option here and you don't need to pass version into `request` down below.

extension/content/components/pages/RecipesPage.js
```diff
@@ -303,12 +345,14 @@ export default function WrappedRecipePage(props) {
...
   const environments = useEnvironments();
   const api = useSelectedEnvironmentAPI();
+  const auth = useSelectedEnvironmentAuth();
   return (
     <RecipesPage
       {...props}
       environmentKey={selectedKey}
       environment={environment}
       environments={environments}
+      auth={auth}
```
rehandalal: it doesn't seem like `auth` is required

extension/content/components/pages/RecipesPage.js
```diff
@@ -303,12 +345,14 @@ export default function WrappedRecipePage(props) {
   const environment = useSelectedEnvironment();
   const environments = useEnvironments();
   const api = useSelectedEnvironmentAPI();
+  const auth = useSelectedEnvironmentAuth();
```
rehandalal: remove if auth is not required below

extension/content/components/pages/RecipesPage.js
```diff
@@ -21,7 +23,9 @@ import {
   useSelectedEnvironment,
   useSelectedEnvironmentAPI,
 } from "devtools/contexts/environment";
+import RecipeEditor from "devtools/components/recipes/RecipeEditor";
 import { convertToV1Recipe } from "devtools/utils/recipes";
+import { useSelectedEnvironmentAuth } from "../../contexts/environment";
```
rehandalal: i'm guessing we can get rid of this import, but if not it should move up to where the other imports from this file happen and we should never use relative path imports

extension/content/components/recipes/RecipeListing.js
```diff
@@ -2,7 +2,10 @@ import autobind from "autobind-decorator";
 import yaml from "js-yaml";
 import PropTypes from "prop-types";
 import React from "react";
-import { Button, Icon, Panel, Tag } from "rsuite";
+
```
rehandalal: nit: remove new line

extension/content/components/recipes/RecipeListing.js
```diff
@@ -2,7 +2,10 @@ import autobind from "autobind-decorator";
...
+import { Link } from "react-router-dom";
+
```
rehandalal: nit: remove new line

extension/content/components/recipes/RecipeListing.js
```diff
@@ -2,7 +2,10 @@ import autobind from "autobind-decorator";
...
+import { Button, Icon, Panel, Tag, ButtonToolbar } from "rsuite";
 import { convertToV1Recipe } from "devtools/utils/recipes";
```
rehandalal, line 8: nit: add new line before this.

the rationale here is you are grouping library/framework imports and project imports

tests/conftests.js
```diff
@@ -1,8 +1,23 @@
 /* eslint-env node */
 import crypto from "crypto";
-
 Object.defineProperty(global.self, "crypto", {
```
rehandalal: nit: new line before this.

tests/conftests.js
```diff
@@ -1,8 +1,23 @@
...
   value: {
     getRandomValues: arr => crypto.randomBytes(arr.length),
   },
 });
+browser.experiments.normandy = {
```
rehandalal, line 10: nit: new line before this.

tests/conftests.js
```diff
@@ -1,8 +1,23 @@
...
+  checkRecipeFilter: () => null,
+  runRecipe: () => null,
+};
+browser.identity = {
```
rehandalal, line 15: nit: new line before this

tests/conftests.js
```diff
@@ -1,8 +1,23 @@
...
+  getRedirectURL: () => {},
+  launchWebAuthFlow: () => {
+    return "https://08e23a90be40fa842a9f18ac049d45a7c3c8e7ec.extensions.allizom.org/#access_token=abc&scope=openid%20profile%20email&expires_in=7200&token_type=Bearer&state=abc&id_token=abc";
+  },
+};
+global.document.body.createTextRange = () => ({
```
rehandalal, line 22: nit: new line before this

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,193 @@
...
+    let recipe = await api.fetchRecipe(id, 3);
+    const { latest_revision } = recipe;
+    latest_revision.arguments = JSON.stringify(
+      latest_revision.arguments,
+      null,
+      1,
+    );
+    latest_revision.action_id = latest_revision.action.id;
+    delete latest_revision.comment;
+    setData(latest_revision);
```
rehandalal: i think this should just be:
```js
const recipe = await api.fetchRecipe(id);
setData(recipe.latest_revision)
```

any manipulation of data should happen at the field level, not here

also you don't need the version passed through on `fetchRecipe` since it is using v3.

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,193 @@
...
+      data: requestBody,
+    });
+    requestSave
+      .then(() => {
+        location.replace("/content.html#");
+        Alert.success("Changes Saved");
+      })
+      .catch(err => {
+        Alert.error(`An Error Occurred: ${JSON.stringify(err.data)}`);
+      });
```
rehandalal: this should be a new method or a few new methods on `NormandyAPI`. Let's have a quick chat about this on Friday.

rehandalal (changes requested):
so a few more things:
- It seems like all the changes to `RecipesPage.js` can probably be undone?
- You should delete `utils/request.js` since that's already covered in `utils/api.js`.

extension/content/components/pages/RecipesPage.js
```diff
@@ -254,13 +257,20 @@ class RecipesPage extends React.PureComponent {
...
+    const { page, count } = this.state;
+    const { match } = this.props;
     return (
       <React.Fragment>
         <Header>
           <Navbar>
             <Nav pullRight>
+              <Nav.Item
+                componentClass={Link}
+                to={`${match.path}/new`}
```
rehandalal: this seems unnecessarily complicated instead of just doing `to="/recipes/new"`.

extension/content/components/recipes/RecipeListing.js
```diff
@@ -149,8 +152,18 @@ class RecipeListing extends React.PureComponent {
...
   }
 
-  renderViewRecipeButton() {
-    return <Button onClick={this.handleshowRecipeButton}>View Recipe</Button>;
+  renderRecipeButtonToolBar() {
+    const { match, recipe } = this.props;
+
+    return (
+      <ButtonToolbar>
+        <Button componentClass={Link} to={`${match.path}/edit/${recipe.id}`}>
```
rehandalal: I think here too you can just go with ```to={`/recipes/edit/${recipe.id}`}```

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,173 @@
...
+  }, []);
+
+  const handleActionIDChange = value => {
+    if (data.action) {
+      data.action.id = value;
+      setData({ ...data });
+    } else {
+      data.action = { id: value };
+      setData({ ...data });
+    }
```
rehandalal: you are actually mutating `data` here which is not ideal. this could simply be:
```js
setData({
  ...data,
  action: {
    ...data.action,
    id: value,
  },
});
```

besides the benefit of not causing a mutation to the data object you also don't need this `if` block since you can spread `undefined` and it will simply leave the new object empty

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,173 @@
...
+            placeholder={"Select an action"}
+            data={actions}
+            block
+            value={data.action ? data.action.id : null}
+            onChange={value => handleActionIDChange(value)}
+          />
+        </FormGroup>
+        <ActionArgument
+          name="arguments"
+          value={JSON.stringify(data.arguments, null, 1)}
```
rehandalal: i'm not sure if you tested this but you can no longer edit the action arguments. this is because we are using a controlled version of codemirror. for this use case we will likely need to use the uncontrolled version and it will need to be handled appropriately at the time of submission

rehandalal (changes requested):
nice job! we're so close! just a couple more things.

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,173 @@
...
+          }}
+          value={props.value}
+          onBeforeChange={(editor, data, value) =>
+            props.handleChange(props.name, JSON.parse(value))
+          }
+        />
+      </FormGroup>
+    );
+  }
+  return "";
```
rehandalal: nit: always return `null` when no component is expected.

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,173 @@
...
+  return (
+    <div className="page-wrapper">
+      <Form fluid formValue={data} onChange={data => setData(data)}>
+        <FormGroup>
+          <ControlLabel>Name</ControlLabel>
+          <FormControl name="name" data-testid="recipeName" />
+          <HelpBlock>Required</HelpBlock>
+        </FormGroup>
+        <FormGroup>
+          <ControlLabel>Experiment Slug</ControlLabel>
```
rehandalal: nit: Experiment**er** Slug

rehandalal (changes requested):
couple other little nits

package.json
```diff
@@ -18,7 +18,7 @@
     "lint:eslint-fix": "yarn run lint:eslint --fix",
     "test": "jest",
     "watch": "npm-run-all build:webpack-dev --parallel --race watch:*",
-    "watch:extension": "web-ext run -s dist",
+    "watch:extension": "web-ext run -s dist --firefox=\"/Applications/Firefox Nightly.app/Contents/MacOS/firefox-bin\"",
```
rehandalal: oops!

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,190 @@
...
+import {
+  UnControlled as CodeMirrorUC,
+  Controlled as CodeMirror,
+} from "react-codemirror2";
+import { useSelectedEnvironmentAPI } from "devtools/contexts/environment";
+
+export default function RecipeEditor(props) {
+  const { match } = props;
+  const api = useSelectedEnvironmentAPI();
+  let [data, setData] = useState({});
```
rehandalal: nit: should be `const`

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,190 @@
...
+  let [actions, setActions] = useState([]);
```
rehandalal: nit: should be `const`

extension/content/components/recipes/RecipeEditor.js
rehandalal: also maybe update the initial state value to:
```js
{
  arguments: {},
}
```

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,190 @@
...
+
+  async function getActionsOptions() {
+    let res = await api.fetchActions(3);
```
rehandalal: nit: should be `const`

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,190 @@
...
+    let actions = res.results.map(action => ({
```
rehandalal: nit: should be `const`

extension/content/utils/api.js
```diff
@@ -91,10 +99,29 @@ export default class NormandyAPI {
...
-  async fetchRecipe(id, { version = 3 }) {
+  async fetchRecipe(id) {
     return this.request({
       url: `recipe/${id}/`,
-      version,
+    });
+  }
+
+  async saveRecipe(id, data) {
+    let url = "recipe/";
```
rehandalal, line 109: nit: should be `const`

extension/content/utils/api.js
```diff
@@ -91,10 +99,29 @@ export default class NormandyAPI {
...
+    let method = "POST";
```
rehandalal, line 110: nit: should be `const`

extension/content/components/recipes/RecipeEditor.js
rehandalal: no version necessary there is only a version 3 endpoint

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,190 @@
...
+      label: action.name,
+      value: action.id,
+    }));
+
+    setActions(actions);
+  }
+
+  async function getRecipe(id) {
+    const recipe = await api.fetchRecipe(id, 3);
```
rehandalal: no version necessary

rehandalal (changes requested):
some last little nits and we should be good to merge this

extension/content/utils/api.js
tiftran: doesn't that make url not happy when it changes to "recipe/< some id>" below?

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,190 @@
...
+            placeholder={"Select an action"}
+            data={actions}
+            block
+            value={data.action ? data.action.id : null}
+            onChange={value => handleActionIDChange(value)}
+          />
+        </FormGroup>
+        <ActionArgument
+          name="arguments"
+          value={JSON.stringify(data.arguments, null, 1)}
```
rehandalal: oops sorry one more, this should be indented `2` not `1`

rehandalal (commented):
one more nit, but also in testing it seems like the `onChange` may be breaking things

tiftran:
I changed up onChange, but it's kinda clunky, if you continuously edit the cursor always moves to the end

rehandalal:
yeah i was hoping this `onChange` solution would work but I think the best thing to do is simply handle this like a true uncontrolled input and pull its value at submission time. to make this work with a functional component you will need to forward the ref.

Here are some guidelines:
https://reactjs.org/docs/forwarding-refs.html
https://reactjs.org/docs/uncontrolled-components.html

Let me know if you want to chat about this

extension/content/components/recipes/RecipeEditor.js
```diff
@@ -0,0 +1,189 @@
...
+        Alert.success("Changes Saved");
+      })
+      .catch(err => {
+        Alert.error(`An Error Occurred: ${JSON.stringify(err.data)}`, 5000);
+      });
+  };
+
+  const formatRequestBody = () => {
+    /* eslint-disable no-unused-vars */
+    const { ["comment"]: _omitData, ...requestBody } = data;
```
rehandalal: this should be:
```js
const { comment: _omitComment, ...cleanedData } = data;
```

primarily the change to remove the square braces but I think those names are also a little clearer

also maybe include a TODO about removing this once https://github.com/mozilla/normandy/issues/2189 is addressed

rehandalal: After talking to @mythmon we've decided we should keep the comment field so I'll file an issue for how to deal with it. no TODO required.

rehandalal (approved):
👏🏽

bors r+

PR #79, opened 2020-04-09
Add environment to URL and copy functionality to address bar

### LINKED ISSUE #52: Wire up the copy button on the address bar
In #51 we updated the NDT address bar. It now has a copy button that needs to be wired up:

image

rehandalal:
This adds the selected environment to the URL which allows us to link to specific environments when linking to NDT.

Also:
Fixes #52 

r?


tiftran (approved):
looks good to me

PR #93, opened 2020-04-22
Refactor of the recipe editing page

rehandalal:
There a bunch of little usability tweaks in here but the bulk of this was to facilitate converting the recipe form to having it's own context so fields can be split out from the form component more easily and there's less need for deep prop drilling.

r?


extension/content/components/pages/RecipeFormPage.js
```diff
@@ -0,0 +1,53 @@
...
+      });
+    } else if (experimenterSlug) {
+      experimenterApi
+        .fetchRecipe(experimenterSlug)
+        .then(({ comment, ...recipeData }) => {
+          setData(recipeData);
+          setImportInstructions(comment);
+        });
+    } else {
+      setData({});
```
tiftran: was fiddling with it and the create workflow wasnt working for me until I inserted filter_object and arguments here...
not exactly sure if this is the right place to do that though...

tiftran (commented):
Thank you for making this some much more clean 🙏 

I tried experimenter -> ndt, experimenter sends action name which I guess since ndt is using action_id at the moment it didn't translate over?

rehandalal:
@tiftran addressed the issues you mentioned and also made a small change to the name of the context since I think we will also use this context for the recipe details page.

tiftran (approved):

PR #95, opened 2020-04-22
Use -f nightly flag with web-ext

rehandalal:
r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,12 @@ $ yarn watch:webpack
 Second, automatically rebuild the XPI file and update it in a Nightly temporary profile
 
 ```
-$ yarn watch:extension --firefox path/to/nightly
+$ yarn watch:extension
 ```
+
+> NOTE: If you are seeing an error while trying to run `watch` make sure that Nightly
+> is installed on your system and if you are on Linux make sure that `firefox-nightly`
+> is on your PATH.
 
 # Release
 
--- a/README.md
+++ b/README.md
@@ -83,18 +83,11 @@ $ cd normandy-devtools
 $ yarn install
 ```
 
-To run the extension in development mode, you'll need two commands in two separate terminals:
-
-First automatically rebuild the extension files
-
-```
-$ yarn watch:webpack
-```
-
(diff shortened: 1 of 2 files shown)


mythmon (approved):
This is correct, but now we don't need to run the two commands in separate terminals, as `yarn watch` should just work. I'm not sure if this a better workflow or not (I haven't tried it). It might be worth only mentioning `yarn watch`, or softening the language of running the command in separate terminals.

Or we could leave it as it is now, your call.

bors delegate+

rehandalal:
I only use the `yarn watch` workflow and it works well for me. Although I think there's room for a little bit of improvement because it does a double build to start which is a little annoying.

rehandalal (commented):
bors r=@mythmon

mythmon:
> I only use the `yarn watch` workflow and it works well for me. Although I think there's room for a little bit of improvement because it does a double build to start which is a little annoying.

That's actually by design. `webpack --watch` will build do a build first, and then start watching. There isn't a way to wait for that first build to finish though, and we need to make sure there is some build before starting web-ext. So `yarn watch` runs a normal build first, and then starts both webpack and web-ext in watch mode in parallel.

There has been [an open issue for Webpack](https://github.com/webpack/webpack/issues/4991) for a long time to give it an option to watch but not build, and it would solve this exact case. There is even a PR for it, but it hasn't been merged.

Is the double build actually slowing you down, or just wasteful? I wonder if we could make the second build basically a no-op if we tweak the caching or some other options.

rehandalal:
It's not really slowing me down at most times. It's just if I'm making certain kind of changes like to the webpack config where I need to restart it can sometimes be annoying. I'm not too concerned about it.

mythmon:
Ah, ok. The second build doesn't block the extension starting, it's just noise. The first build, which is blocking in your case, is needed anyways. And watch mode rebuilds seem pretty quick to me. So the double build isn't making anything worse.

It sounds like there might be some small benefit in trying to optimize our build times, but it's not a big deal.

PR #97, opened 2020-04-22
Speed up webpack by using a faster devtool and caching more

mythmon:
This makes dev rebuilds 5x faster on my system, without affecting prod builds.

To test:

0. pull this branch
1. `yarn install` to pick up new deps
2. `yarn build:webpack-dev` to prime the cache (this would be done automatically with yarn watch, but lets make it explicit)
3. `yarn watch` and notice how much faster things are

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules/
 web-ext-artifacts/
 yarn-error.log
 .env
+.webpack-cache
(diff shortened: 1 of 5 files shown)


rehandalal (approved):
this is awesome!!! thanks @mythmon!

bors r+

rehandalal:
i also had a 5x improvement.

PR #98, opened 2020-04-23
Tweak webpack.config.js

rehandalal:
couple little tweaks. the default for `argv` is to fix an issue my IDE was reporting:

> Can't analyse webpack.config.js: coding assistance will ignore module resolution rules in this file. Possible reasons: this file is not a valid webpack configuration file or its format is not currently supported by the IDE. Error details:  Cannot read property 'mode' of undefined

r?


mythmon (approved):
bors r+

PR #106, opened 2020-04-23
added heartbeat action args

### LINKED ISSUE #57: Add action arguments fields for show-heartbeat action
Depends on #54 

Existing UI in DC:
image

tiftran:
fixes #57


extension/content/components/recipes/form/arguments/fields/CheckboxField.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+    });
+  };
+
+  return (
+    <FormGroup>
+      <ControlLabel>{label}</ControlLabel>
+      <Checkbox
+        checked={data.arguments[name]}
+        onChange={(_, value) => handleChange(value)}
```
rehandalal: instead of creating another anonymous function here we should just update the arguments to `handleChange` above.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+} from "devtools/contexts/recipeDetails";
+
+import InputField from "devtools/components/recipes/form/arguments/fields/InputField";
+import SelectField from "devtools/components/recipes/form/arguments/fields/SelectField";
+import CheckboxField from "devtools/components/recipes/form/arguments/fields/CheckboxField";
+import InputUrlField from "devtools/components/recipes/form/arguments/fields/InputUrlField";
+
+export default function ShowHeartBeatArguments() {
+  return (
+    <Grid fluid>
```
rehandalal: This is causing the inputs to become misaligned. I would replace it with a `FormGroup` instead.

extension/content/components/recipes/form/arguments/fields/InputUrlField.js
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <FormGroup>
+      <ControlLabel>{label}</ControlLabel>
+      <InputGroup>
+        <Input value={data.arguments[name]} onChange={handleChange} />
+        <InputGroup.Button onClick={() => window.open(data.arguments[name])}>
+          <Icon icon="link" />
+          View
+        </InputGroup.Button>
+      </InputGroup>
```
rehandalal: I know this is in the original UI but I don't think there's a lot of value having a link out from the form.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+  const dispatch = useRecipeDetailsDispatch();
+
+  const handleChange = (name, value) => {
+    dispatch({
+      type: ACTION_UPDATE_DATA,
+      data: {
+        ...data,
+        arguments: {
+          ...data.arguments,
+          [name]: parseInt(value),
```
rehandalal: unfortunately this breaks things when you clear the field. see here:
https://github.com/mozilla-extensions/normandy-devtools/blob/master/extension/content/components/recipes/form/arguments/PreferenceExperimentArguments.js#L216

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+          <RepeatOptionsField />
+        </Col>
+        <Col xs={12}>
+          <InputField
+            label="Engagement Button Label"
+            name="engagementButtonLabel"
+          />
+          <InputField label="Thanks Message" name="thanksMessage" />
+          <InputUrlField label="Post-Answer URL" name="postAnswerUrl" />
+          <CheckboxField
```
rehandalal: I think you could just use `ToggleField` for this.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+          <InputField label="Learn More Message" name="learnMoreMessage" />
+          <RepeatOptionsField />
+        </Col>
+        <Col xs={12}>
+          <InputField
+            label="Engagement Button Label"
+            name="engagementButtonLabel"
+          />
+          <InputField label="Thanks Message" name="thanksMessage" />
+          <InputUrlField label="Post-Answer URL" name="postAnswerUrl" />
```
rehandalal: I think an `InputField` should be fine for this based on my comments below about not needing the ability to link out from the field.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+        </FormGroup>
+      );
+    }
+
+    return null;
+  };
+
+  return (
+    <React.Fragment>
+      <SelectField
```
rehandalal: This probably needs a side effect to remove the "repeatsEvery" value from the data unless that is the option selected.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+        "Show until the user interacts with the heartbeat prompt, then never again.",
+      value: "nag",
+    },
+  ];
+
+  const data = useRecipeDetailsData();
+  const dispatch = useRecipeDetailsDispatch();
+
+  const handleChange = (name, value) => {
+    dispatch({
```
rehandalal: a pattern i used for this kind of thing was to have `handleChange` return the handler function:
https://github.com/mozilla-extensions/normandy-devtools/blob/master/extension/content/components/recipes/form/arguments/PreferenceExperimentArguments.js#L195

you wouldn't need the callback (cb) for this.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+        arguments: {
+          ...data.arguments,
+          [name]: parseInt(value),
+        },
+      },
+    });
+  };
+
+  const repeatEveryOption = () => {
+    if (data.arguments.repeatOption === "xdays") {
```
rehandalal: nit: i would do:
```js
if (data.arguments.repeatOption !== "xdays") {
  return null;
}
```

this leaves the your JSX one level less indented.

extension/content/components/recipes/form/arguments/ShowHeartBeatArguments.js
```diff
@@ -0,0 +1,118 @@
...
+      {repeatEveryOption()}
+    </React.Fragment>
+  );
+}
+
+ShowHeartBeatArguments.propTypes = {
+  handleChange: PropTypes.func,
+  name: PropTypes.name,
+  value: PropTypes.string,
+};
```
rehandalal: you don't need these

rehandalal (changes requested):
👏🏽 nice work!

had a few comments and some things that need to be fixed.

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -0,0 +1,135 @@
...
+          value={data.arguments.repeatEvery}
+          onChange={handleChange("repeatEvery")}
+        />
+        <HelpBlock>Required</HelpBlock>
+      </FormGroup>
+    );
+  };
+
+  return (
+    <React.Fragment>
```
rehandalal: nit: I've been using the shorthand `<>...</>` instead of `React.Fragment`.

rehandalal: i'll file a PR with an eslint rule to push for consistency

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -0,0 +1,135 @@
...
+        <InputNumber
+          min={1}
+          postfix="days"
+          value={data.arguments.repeatEvery}
+          onChange={handleChange("repeatEvery")}
+        />
+        <HelpBlock>Required</HelpBlock>
+      </FormGroup>
+    );
+  };
```
rehandalal: i think you can just split this out into its own function below:
```js
function RepeatEveryField() {
  const data = useRecipeDetailsData();

  if (data.arguments.repeatOption !== "xdays") {
    return null;
  }
  ...
```

rehandalal: also i think the way you've set up this field it switches between controlled and uncontrolled and that's causing issues. you will need to make sure that `data.arguments.repeatEvery` has a numeric value or an empty string value before this field is rendered

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -0,0 +1,135 @@
...
+      value: "xdays",
+    },
+    {
+      label:
+        "Show until the user interacts with the heartbeat prompt, then never again.",
+      value: "nag",
+    },
+  ];
+
+  const repeatsEveryChangeSideEffect = ({ data, value }) => {
```
rehandalal: nit: this should be `repeatOptionChangeSideEffect` since `repeatOption` is the field that is changing.

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -0,0 +1,135 @@
...
+    if (value !== "xdays") {
+      /* eslint-disable no-unused-vars */
+      const {
+        repeatEvery: _omitRepeatOption,
```
rehandalal: nit: `_omitRepeatEvery` or just `_omit`

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -0,0 +1,135 @@
...
+        action,
+        ...cleanedArguments
+      } = data.arguments;
+      /* eslint-enable no-unused-vars */
+      return { ...data, arguments: cleanedArguments };
+    }
```
rehandalal, line 74: you probably need an else here:
```js
} else {
  return { repeatEvery: 1, ...data };
}
```

that would solve the issue with the `repeatEvery` field flipping between controlled and uncontrolled.

rehandalal (changes requested):
couple other little things to fix up and this should be good to go

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -61,21 +61,35 @@ function RepeatOptionsField() {
...
+        repeatEvery: _omitRepeatEvery,
         action,
         ...cleanedArguments
       } = data.arguments;
       /* eslint-enable no-unused-vars */
       return { ...data, arguments: cleanedArguments };
     }
 
-    return data;
+    return { ...data, arguments: { ...data.arguments, repeatEvery: 1 } };
```
rehandalal: nit: order matters here so it should technically be `{ repeatEvery: 1, ...data.arguments }`

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -61,21 +61,35 @@ function RepeatOptionsField() {
...
+        label="How often should the prompt be shown?"
+        name="repeatOption"
+        options={SHOW_PROMPT_OPTIONS}
+      />
+      {repeatEveryOption()}
+    </>
+  );
+}
+
+function repeatEveryOption() {
```
rehandalal: almost there. this should be a functional component so something like `RepeatEveryField` not `repeatEveryOption`.

extension/content/components/recipes/form/arguments/ShowHeartBeat.js
```diff
@@ -61,21 +61,35 @@ function RepeatOptionsField() {
...
 
+  return (
+    <>
+      <SelectField
+        changeSideEffect={repeatOptionChangeSideEffect}
+        label="How often should the prompt be shown?"
+        name="repeatOption"
+        options={SHOW_PROMPT_OPTIONS}
+      />
+      {repeatEveryOption()}
```
rehandalal: used here as 
```js
<RepeatEveryField />
```

rehandalal (changes requested):

rehandalal (approved):
thanks!!!

bors r+

PR #107, opened 2020-04-23
Bump to v2.0.0-alpha.1

rehandalal:
Prep for an alpha release

r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "normandy-devtools",
-  "version": "1.0.0",
+  "version": "2.0.0-alpha.1",
   "description": "A set of tools for interacting with the Firefox Normandy client",
   "homepage_url": "https://github.com/mozilla/normandy-devtools",
   "private": true,


mythmon:
I was able to build this, and it all seems to work as intended so far. One bug I found is that after editing a stage recipe, I was redirected back to the prod recipe listing. I think that signing in to Dev also didn't update my avatar until after I switched environments. Stage worked fine in that regard though.

There are two major things that I don't think we can do yet that should block the final release: There is no way to review/approve recipes, or to enable/disable them. I expect this would show up on a recipe detail page.

I think this is something that we can tag as the first alpha of 2.0. There is definitely more work to be done, but I think we could start using this for some launch tasks, especially simple ones like pauses. I think this is something that we could have people start poking at it, and filing bugs if they can. I'll file bugs about any issues that I see, including the ones I mentioned above.

mythmon (approved):

rehandalal:
bors r=@mythmon

PR #108, opened 2020-04-24
Add arguments fields for message-experiment action

### LINKED ISSUE #88: Add action arguments fields for messaging-experiments action
We need custom action arguments UI for "messaging experiments".

Action schema:
https://searchfox.org/mozilla-central/source/toolkit/components/normandy/actions/schemas/index.js#22

rehandalal:
Fixes #88 

image

r?


extension/content/components/recipes/form/arguments/MessagingExperiment.js
```diff
@@ -0,0 +1,214 @@
...
+  const parseNumericInput = (value) => {
+    if (!value) {
+      return "";
+    }
+
+    return parseInt(value, 10);
+  };
+
+  let valueFieldKey = data.recipe && data.recipe.id ? data.recipe.id : "create";
+  valueFieldKey += `-${index}-${branchesUpdated}`;
```
rehandalal, line 153: this was a tricky thing to sort out but makes sure that when you delete a branch the value json editor field updates correctly

extension/content/components/recipes/form/arguments/MessagingExperiment.js
```diff
@@ -0,0 +1,221 @@
...
+
+function Branches() {
+  const [branchesUpdated, setBranchesUpdated] = React.useState(Date.now());
+  const data = useRecipeDetailsData();
+  const dispatch = useRecipeDetailsDispatch();
+  const { preferenceType } = data.arguments;
+
+  const handleClickAddBranch = () => {
+    const newBranch = { ratio: 1, slug: "", value: {}, groups: [] };
+    if (preferenceType === "boolean") {
```
tiftran: Probs a silly question, but is pref type supposed to be an editable field somewhere?

tiftran (commented):
A silly question and I also noticed that after saving and going back to edit, the group values that were saved don't display.

extension/content/components/recipes/form/arguments/MessagingExperiment.js
rehandalal: oops! this should not be there at all. that's what i get for making a bowl of copy pasta.

rehandalal: nice catch tho!

extension/content/components/recipes/form/arguments/MessagingExperiment.js
```diff
@@ -0,0 +1,221 @@
+// @ts-nocheck
```
mythmon, line 1: Out of curiosity, did you add this as copy/paste, or because you ran into some trouble and didn't want to deal with it right now?

extension/content/components/recipes/form/arguments/MessagingExperiment.js
```diff
@@ -0,0 +1,221 @@
...
+            }),
+          },
+        },
+      });
+    };
+  };
+
+  const parseNumericInput = (value) => {
+    if (!value) {
+      return "";
```
mythmon, line 146: So this returns a number sometimes and a string other times? Should it be `null` instead of `""`?

mythmon (approved):
I tested this out be creating a recipe on stage, and it seems to work. I was able to make and edit a recipe, and nothing exploded. :+1:

extension/content/components/recipes/form/arguments/MessagingExperiment.js
rehandalal: `null` causes issues where it switches between controlled and uncontrolled for some reason. i believe that's related to how rsuite passes value through

extension/content/components/recipes/form/arguments/MessagingExperiment.js
rehandalal: the latter

extension/content/components/recipes/form/arguments/MessagingExperiment.js
mythmon: Ah, I can see how that happens. Thanks for the explanation!

extension/content/components/recipes/form/arguments/MessagingExperiment.js
rehandalal: i added ts-nocheck to the files that the check complained about

mythmon: Ah, this points out an interesting thing. I ran the type checker on this to see what it was complaining about. It's reading the actual parameters of the functional component, and ignoring the proptypes. The function parameters make it look like `options` is required (since it doesn't have a default value), but the proptypes say it's not. I'm sure this is going to cause some confusion. Oh well. This is fine for now.

tiftran (approved):

PR #112, opened 2020-04-24
Add Typescript configured to let us gradually convert

mythmon:

PR #121, opened 2020-04-27
Restore page

mythmon:
Talking with Rehan, this seems like it would be a nice convenient change to the dev workflow.

All the new files are written in Typescript, so this depends on #112. If we end up not merging #112, I can redo this in plan JS.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -1,7 +1,11 @@
-browser.experiments.normandy.setManualMode(true);
-
 browser.browserAction.onClicked.addListener(async () => {
   await browser.tabs.create({
     url: "content.html",
   });
 });
+
+if (DEVELOPMENT) {
+  browser.tabs.create({
+    url: "restore.html",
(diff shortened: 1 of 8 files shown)


extension/background.js
```diff
@@ -1,7 +1,11 @@
...
   await browser.tabs.create({
     url: "content.html",
   });
 });
+
+if (DEVELOPMENT) {
+  browser.tabs.create({
+    url: "restore.html",
+  });
+}
```
rehandalal, line 11: I wonder if we wanted finer grained control over this through an environment variable like:
```sh
NDT_RESTORE=last-url
```

i'm thinking about the future where i want to have it restore all the tabs you had opened

rehandalal: But you know what I can deal with that in a future PR

rehandalal (approved):
this works great, thanks!

bors r+

PR #122, opened 2020-04-28
Run tests on CI and use node 14.0.0 for builds

rehandalal:
I updated the wrong job and realized that job isn't even being run.

r?


package.json
```diff
@@ -12,10 +12,9 @@
     "build:webpack-dev": "webpack --mode development --progress --hide-modules",
     "build:finalize": "mv web-ext-artifacts/*.zip web-ext-artifacts/normandy-devtools.zip",
     "clean": "rimraf dist web-ext-artifacts .webpack-cache",
-    "lint": "npm-run-all lint:eslint",
-    "lint-fix": "npm-run-all lint:eslint-fix",
-    "lint:eslint": "eslint webpack.config.js extension",
-    "lint:eslint-fix": "yarn run lint:eslint --fix",
+    "lint": "npm-run-all eslint stylelint",
+    "eslint": "eslint webpack.config.js extension",
+    "stylelint": "stylelint extension/**/*.less extension/**/*.css",
```
mythmon: Can we keep the `group:specific` pattern here, and name these `lint:eslint` and `lint:stylelint`? Then we could make the lint job `npm-run-all lint:*`.

mythmon (approved):

package.json
rehandalal: doing that causes TC to run each of those individually which I was trying to avoid since it looks for any scripts that start with `lint`. we could use another prefix like `check`? and do `check:*`?

rehandalal:
`lint` is now just an alias for `check`

package.json
mythmon: :frowning:

mythmon (approved):
bors merge

PR #124, opened 2020-04-28
pref rollback args

### LINKED ISSUE #61: Add action arguments fields for preference-rollback action
Depends on #54.

Existing UI in DC:
image

tiftran:
fixes #61

### DIFF
--- a/extension/content/utils/normandyApi.js
+++ b/extension/content/utils/normandyApi.js
@@ -51,7 +51,7 @@ export default class NormandyAPI extends API {
   async fetchAllRecipes(searchParams = {}) {
     let response = await this.request({
       url: "recipe/",
-      data: { searchParams },
+      data: searchParams,
     });
     let recipes = response.results;
 
--- a/extension/content/utils/normandyApi.js
+++ b/extension/content/utils/normandyApi.js
@@ -48,10 +48,10 @@ export default class NormandyAPI extends API {
     });
   }
 
-  async fetchRecipeSearchParams(searchParams = {}) {
+  async fetchAllRecipes(searchParams = {}) {
     let response = await this.request({
       url: "recipe/",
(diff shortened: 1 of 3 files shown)


extension/content/utils/normandyApi.js
```diff
@@ -48,6 +48,23 @@ export default class NormandyAPI extends API {
     });
   }
 
+  async fetchRecipeSearchParams(searchParams = {}) {
```
rehandalal: nit: rename to `fetchAllRecipes` so that it's clear we're fetching all the recipes and it's a little clearer why we're not just using `fetchRecipePage`.

extension/content/utils/normandyApi.js
```diff
@@ -48,6 +48,23 @@ export default class NormandyAPI extends API {
...
+    let response = await this.request({
+      url: "recipe/",
+      data: { ...searchParams },
```
rehandalal: should be: 
```js
data: searchParams,
```

rehandalal (changes requested):
nice!

just a couple little nits and we can 🚢 this!

extension/content/utils/normandyApi.js
```diff
@@ -48,10 +48,10 @@ export default class NormandyAPI extends API {
     });
   }
 
-  async fetchRecipeSearchParams(searchParams = {}) {
+  async fetchAllRecipes(searchParams = {}) {
     let response = await this.request({
       url: "recipe/",
-      data: { ...searchParams },
+      data: { searchParams },
```
rehandalal: this should not be wrapped in `{}`. it should just be:
`data: searchParams,`

rehandalal (changes requested):

rehandalal (approved):
bors r+

PR #128, opened 2020-04-28
Stop using check as a prefix and go back to lint

rehandalal:
r?


mythmon (approved):
I'm ok with this change. Lets make sure that Taskcluster likes it too though, since we still aren't checking that with bors.

mythmon:
bors merge

PR #131, opened 2020-04-29
remove default version value from api.js

### LINKED ISSUE #130: experimenter import endpoint url uses incorrect version
Screen Shot 2020-04-28 at 8 40 05 PM

tiftran:
fixes #130 
So... I took a stab at this....

removed the version value default since the experimenter api uses version value of 1?

### DIFF
--- a/extension/content/utils/api.js
+++ b/extension/content/utils/api.js
@@ -8,6 +8,13 @@ export default class API {
     throw new Error("getBaseURL() needs to be implemented");
   }
 
+  /**
+   * @param {Object} args
+   * @param {string} args.url
+   * @param {number} [args.version]
+   * @param {Object} [args.extraHeaders]
+   * @param {string|FormData} [args.body]
+   */
   async request({ url, version, extraHeaders = {}, ...options }) {
     const headers = new Headers();
     headers.append("Accept", "application/json");
--- a/extension/content/utils/api.js
+++ b/extension/content/utils/api.js
@@ -8,7 +8,7 @@ export default class API {
     throw new Error("getBaseURL() needs to be implemented");
   }
 
-  async request({ url, extraHeaders = {}, version = 3, ...options }) {
+  async request({ url, version, extraHeaders = {}, ...options }) {
     const headers = new Headers();
     headers.append("Accept", "application/json");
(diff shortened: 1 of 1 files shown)


tiftran:
oh no, i guess i made typescript unhappy...

rehandalal (approved):
good catch!

bors r+

extension/content/utils/api.js
```diff
@@ -8,7 +8,7 @@ export default class API {
     throw new Error("getBaseURL() needs to be implemented");
   }
 
-  async request({ url, extraHeaders = {}, version = 3, ...options }) {
+  async request({ url, version, extraHeaders = {}, ...options }) {
```
mythmon, line 11: Typescript is grumpy because you removed the default value here, but the Experimenter call site at `extension/content/utils/experimenterApi.js` doesn't pass. It's not effectively required since the ExperimenterAPI overrides `getBaseUrl` and provides a default value their, but Typescript can't know that (and I didn't when I read it).

One way you could fix this is to sort out whether this should be required or not. You could declare it as specifically optional like this:

```suggestion
  /**
   * @param {Object} args
   * @param {string} args.url
   * @param {?number} args.version
   * @param {?Object} args.extraHeaders
   * @param {?string} args.body
   */
  async request({ url, version, extraHeaders = {}, ...options }) {
```

mythmon, line 18: Typescript is grumpy because you removed the default value here, but the Experimenter call site at `extension/content/utils/experimenterApi.js` doesn't pass a value for version. It's not effectively required since the ExperimenterAPI overrides `getBaseUrl` and provides a default value their, but Typescript can't know that (and I didn't when I read it).

One way you could fix this is to sort out whether this should be required or not. You could declare it as specifically optional like this:

```suggestion
  /**
   * @param {Object} args
   * @param {string} args.url
   * @param {number} [args.version]
   * @param {Object} [args.extraHeaders]
   * @param {string|FormData} [args.body]
   */
  async request({ url, version, extraHeaders = {}, ...options }) {
```

tiftran: oooh @mythmon  tyty

PR #137, opened 2020-04-29
Add support for system environment vars

rehandalal:
Allows me to use an `.envrc` file with [direnv](https://direnv.net/) to set all the environment variables.

r?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,5 @@ dist/
 node_modules/
 web-ext-artifacts/
 yarn-error.log
-.env
+.env*
 .webpack-cache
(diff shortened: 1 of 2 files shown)


mythmon:
One concern I have about this is how these environment variables are exposed to the code. If *all* environment variables, even those that aren't used, are included in the build then we have a problem, probably. I bet that Taskcluster exposes secrets through environment variables, and I know that Circle can.

mythmon (approved):
This plugin makes all the variables *available* using Webpack's DefinePlugin, which is doing string replacement. That means that there isn't a JS object somewhere with all of the details. This seems safe to me.

bors merge

PR #138, opened 2020-04-29
Reload when manifest.json or package.json change

### LINKED ISSUE #136: Watch for changes to manifest.json
Currently when we change `manifest.json` we need to stop watching and re-run `yarn watch`. It would be nice for webpack to watch for changes to `manifest.json` and update the manifest when changes occur.

mythmon:
I learned how to make Webpack plugins, and I made one to reload when the manifest.json or package.json changes.

Fixes #136

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -8,9 +8,6 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const webpack = require("webpack");
 const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
 
-const packageData = require("./package.json");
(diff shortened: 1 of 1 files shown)


rehandalal:
this does not appear to work as expected. i haven't dug into it but the build fails because of a missing `manifest.json`.

mythmon:
Weird. This worked fine for me locally, but I did see that kind of error while I was working on it. I'll look into this a little more tomorrow. However, I have another idea for how to deal with this that would avoid the problem we're hitting here entirely.

mythmon:
I don't think this is the right solution anymore.

PR #139, opened 2020-04-29
Add version info to sidebar

mythmon:
Also includes some infrastructure to include built-time constants in the
built output, and to write tests in Typescript.


extension/content/components/common/AppSidebar.tsx
```diff
@@ -0,0 +1,76 @@
+import React from "react";
+import { NavLink } from "react-router-dom";
+import { Icon, Nav, Sidenav, Popover, Whisper } from "rsuite";
+
+import { useEnvironmentState } from "devtools/contexts/environment";
+import { splitCamelCase } from "devtools/utils/helpers";
+
+export default function AppSidebar(): JSX.Element {
```
rehandalal: is there a difference between `JSX.element` and `React.FC`? And should be prefer one or the other?

extension/content/components/common/AppSidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <Nav.Item
+              componentClass={NavLink}
+              icon={<Icon icon="puzzle-piece" />}
+              to={`/${selectedKey}/addon-studies`}
+            >
+              Add-on Studies
+            </Nav.Item>
+          </Nav>
+
+          <div className="version-info flex-grow-0">
```
rehandalal: a couple of these will only land in the details page patch but...

you could use:
`text-subtle cursor-pointer p-2 flex-grow-0`

package.json
```diff
@@ -78,7 +82,7 @@
     "@babel/preset-typescript": "7.9.0",
     "@types/rc-tree": "1.11.3",
     "@types/react-router": "5.1.5",
-    "@types/react-router-dom": "5.1.4",
+    "@types/react-router-dom": "^5.1.5",
```
rehandalal: should we pin these deps?

rehandalal (changes requested):

package.json
mythmon: Oops, yeah. I just added a bunch of type definitions without looking too closely at what was here already.

extension/content/components/common/AppSidebar.tsx
mythmon: I used `JSX.Element` because that is what Typescript inferred the return value of this function was. `JSX.Element` is the type that you get from JSX literals, and specifically, it's the *return value* of this function. `React.FC` (short for `React.FunctionComponent`) is the type of a functional component, and it's the type of this expression:

```
const AppSideBar: React.FC = (props) => { /* some jsx */ }
```

I also learned that `JSX.Element` is an alias for `React.ReactElement` in our project (it would be different in a project that uses JSX but not React).

So I think it depends on how you define the component. If it's `function Foo() {}` then it should be `JSX.Element` or `React.ReactElement`. If it's `const Foo = () => {}`, then it just be `React.FC`.

extension/content/components/common/AppSidebar.tsx
mythmon: I'll use all of these, even though some of them don't work. When your details patch comes through it will get better.

extension/content/components/common/AppSidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+            </Nav.Item>
+          </Nav>
+
+          <div className="text-subtle cursor-pointer p-2 flex-grow-0">
+            <Whisper
+              placement="topStart"
+              speaker={versionPopover}
+              trigger="click"
+            >
+              <button>{__BUILD__.version}</button>
```
rehandalal: i think you should switch this to just a `span` to call less attention to this. i would also move `text-subtle` and `cursor-default` (instead of `cursor-pointer`) to this element instead of the containing `div`.

extension/content/components/common/AppSidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+            >
+              Add-on Studies
+            </Nav.Item>
+          </Nav>
+
+          <div className="text-subtle cursor-pointer p-2 flex-grow-0">
+            <Whisper
+              placement="topStart"
+              speaker={versionPopover}
+              trigger="click"
```
rehandalal: i would just show this on hover

webpack.config.js
```diff
@@ -133,3 +136,16 @@ module.exports = (env, argv = {}) => {
     },
   };
 };
+
+async function getBuildInfo(isDevelopment) {
+  return {
+    isDevelopment,
+    version: (await execOutput("git describe")).trim(),
```
rehandalal: i think this should be:
```js
version: isDevelopment ? (await execOutput("git rev-parse HEAD")).trim() : packageData.version,
```

rehandalal (changes requested):

webpack.config.js
mythmon: If the version building this is a tag, then `git describe` will be just the tag name. The extra info is only included on builds from non-tags. I think it should stay as `git describe`.

extension/content/components/common/AppSidebar.tsx
mythmon: Hover seems too loud, in my opinion. It pops up a big window full of information that isn't often very useful.

extension/content/components/common/AppSidebar.tsx
mythmon: How does making it a button call attention to it? It seems visually the same as a `button` or a `span` to me.

I suppose if it isn't a clickable expansion, and switches to an on-hover, then it makes sense to use span. I'd prefer to keep it as a click interaction though, and so marking it with a button element is more semantic.

mythmon: Moving the utility classes down to the deeper element makes sense though.

rehandalal: > How does making it a button call attention to it? It seems visually the same as a button or a span to me.

Light theme:
image

Dark theme:
image

The button styling comes from the system right now so you might not be seeing it the same way.

As a `span` it looks like this for me (which I believe matches the screenshot you sent me from your system):
image

rehandalal: > I'd prefer to keep it as a click interaction though, and so marking it with a button element is more semantic.

My take on this is if you are providing additional contextual information this functions as a tooltip and therefore I expect it to show on hover. 

If it's a menu like in the case of the avatar in the top right corner when you're signed in and there's a caret icon suggesting dropdown then it could be triggered on click because it's clearer there is a click behavior available.

webpack.config.js
rehandalal: could we add something that complains loudly if the tag name does not match the version in `package.json`? something akin to:
https://github.com/rehandalal/therapist/blob/master/.circleci/config.yml#L103
https://github.com/rehandalal/therapist/blob/master/setup.py#L21

extension/content/components/common/AppSidebar.tsx
mythmon: That makes sense. I thought that the button styling would be coming from some reset somewhere, so I expected it to look the same on all platforms. I don't want to figure out cross platform theming of buttons, so lets go with the span and hover approach.

webpack.config.js
mythmon: Sure.

Fyi, for this project version *should* (and I realize this won't always happen) using `yarn version --minor` (or `--major` or `--patch`). That updates `package.json` and makes a git tag automatically. It's easy enough to enforce that they match though.

rehandalal (approved):
thanks!!!

bors r+

PR #140, opened 2020-04-30
Watch for changes to webpack.config.js

rehandalal:
Sometimes switching between branches where `webpack.config.js` causes things to break and you have to restart watch. This handles that for you. Next up... Watch for `yarn.lock` changes and update dependencies...

r?


mythmon (approved):
That's a clever solution.

bors merge

PR #141, opened 2020-04-30
Add a recipe details page

### LINKED ISSUE #64: Add recipe details page
We need a page where you can see the details of a recipe and view a timeline of changes that allows you to step through old revisions

rehandalal:
Fixes #64 

r?


.eslintrc.js
```diff
@@ -11,6 +11,11 @@ module.exports = {
     "plugin:react/recommended",
   ],
   plugins: ["mozilla"],
+  settings: {
+    react: {
+      version: "detect",
```
mythmon, line 16: Thanks for this, I've been meaning to figure out what that error wanted me to do for a while now.

extension/content/components/recipes/details/CollapsibleSection.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Icon, IconButton } from "rsuite";
+
+export default function CollapsibleSection({
+  children,
+  headerButtons,
+  title,
+  ...props
```
mythmon, line 9: FYI Typescript is going to see these props as required, even though the proptypes below say they're optional.

extension/content/components/common/AppSidebar.tsx
```diff
@@ -13,12 +13,12 @@ export const AppSidebar: React.FC = () => {
       <dl>
         {Object.entries(__BUILD__).map(([key, value]) => {
           return (
-            <>
+            <React.Fragment key={key}>
```
mythmon, line 16: Good call

extension/content/components/recipes/details/ActionDetails.js
```diff
@@ -0,0 +1,110 @@
...
+const MODE_RICH = "RICH";
+const MODE_RAW = "RAW";
+
+const ACTION_DETAILS_MAPPING = {
+  "addon-study": AddonStudy,
+  "branched-addon-study": BranchedAddonStudy,
+  "messaging-experiment": MessagingExperiment,
+  "multi-preference-experiment": MultiPreferenceExperiment,
+  "preference-experiment": PreferenceExperiment,
+  "opt-out-study": AddonStudy,
```
mythmon, line 24: Nit: this list is almost sorted, except out-out-study is out of place.

extension/content/components/recipes/details/ActionDetails.js
```diff
@@ -0,0 +1,110 @@
...
+  "preference-rollout": PreferenceRollout,
+};
+
+export default function ActionDetails() {
+  const data = useRecipeDetailsData();
+  const [mode, setMode] = React.useState(MODE_RICH);
+
+  const handleModeClick = (newMode) => {
```
mythmon: So this is a function that makes a click handler, right? I like the pattern. Can we name it something like `makeHandleModeClick`? Or `handleModeClickFactory` if we're feeling ironic. Especially with the `()=>{}` syntax, I would benefit from some extra hints about what's going on here.

extension/content/components/recipes/details/arguments/AddonStudy.js
```diff
@@ -0,0 +1,29 @@
...
+
+import { useRecipeDetailsData } from "devtools/contexts/recipeDetails";
+import Generic from "devtools/components/recipes/details/arguments/Generic";
+import {
+  nl2pbrFormatter,
+  tagFormatter,
+} from "devtools/components/recipes/details/arguments/formatters";
+
+export default function AddonStudy() {
+  const data = useRecipeDetailsData();
```
mythmon: Why do all the specific action details have to use `useRecipeDetailsData`? It seems like all of the cases where these are used the parent has that data handy. Why not just pass it in?

extension/content/components/recipes/details/arguments/AddonStudy.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <Generic
+      data={data.arguments}
+      formatters={{
```
mythmon, line 14: I really like that the specific views are just calling the Generic one with more formatters.

extension/content/components/recipes/details/arguments/Generic.js
```diff
@@ -0,0 +1,148 @@
+// @ts-nocheck
+import React from "react";
+import PropTypes from "prop-types";
+import { Icon, Tag } from "rsuite";
+import validator from "validator";
+
+import Highlight from "devtools/components/common/Highlight";
+import { splitCamelCase } from "devtools/utils/helpers";
+
+export default function Generic({
```
mythmon: I think that the name `Generic` is too... generic. Maybe all of the argument detail formatters should have `Arguments` on the end? `GenericArugments`, `BranchedAddonStudyArguments`, etc. It duplicated the path a bit, but we see the names in isolation a lot.

extension/content/components/recipes/details/filters/StableSample.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Panel, Tag } from "rsuite";
+
+export default function BucketSample({ data }) {
```
mythmon: This class has the copy/paste name still. Should be `StableSample`

mythmon (changes requested):
There is a lot here. I didn't go through all of it, but the overall shape looks good to me. I have a couple small things I'd like to see different, but I think this is nearly ready to land unless @tiftran has any concerns.

tiftran (commented):
Looks good to me?

Probs not important, but this seems to have appeared...and it doesn't in master
Screen Shot 2020-05-01 at 10 26 33 AM

extension/content/components/recipes/details/filters/StableSample.js
rehandalal: oops! good catch

extension/content/components/recipes/details/CollapsibleSection.js
rehandalal: i figured i'd start doing typescript conversions with the test coverage stuff after next week

mythmon: Oh definitely. I just called this out because Tif ran into this. Since types are also being checked in JS files, it might give confusing errors.

mythmon (approved):
Wooh!

bors merge

PR #146, opened 2020-05-01
Check version matches only on new tags

rehandalal:
this is to prevent issues with bumping the version before tagging so as to be able to land PRs like #107.

r?


mythmon:
Ah, right.

bors merge

PR #147, opened 2020-05-01
added multipref args

### LINKED ISSUE #63: Add action arguments fields for multi-preference-experiment action
Depends on #54

Existing UI in DC:
image

tiftran:
my attempt at multipref 😬 .... 

Fixes #63


mythmon:
When I add two branches, and then delete the first one, the result is that the remaining branch has the name of the second branch but the preferences from the first branch.

tiftran:
so i tried a couple of things... realized it was the branch keys that was cauing the weird rendering thing... not exactly sure if this was the best way to go about it

mythmon:
That seems to have fixed my delete problem.

extension/content/components/recipes/form/arguments/MultiPreferenceBranches.js
```diff
@@ -0,0 +1,293 @@
...
+    if (!value) {
+      return "";
+    }
+
+    return parseInt(value, 10);
+  };
+
+  return (
+    <FormGroup>
+      <div className="d-flex">
```
rehandalal: I think this `div` needs to be wrapped in `FormGroup` to get the appropriate layout margin. As an aside at some point we should look at ditching all the `FormGroup`s for CSS classes that give us the margins we want.

rehandalal (approved):
This generally looks good to me, thanks for all the work on this! 

One little nit and you'll need to merge master in. r+wc

bors d=@tiftran

PR #152, opened 2020-05-01
Use version from package.json in non-development

mythmon:
Taskcluster doesn't build from tags, so if we pull versions from tags we don't get the right thing. This should fix that for now.


rehandalal (approved):
bors r+

PR #157, opened 2020-05-02
Use Fira Code for more consistent monospace font styling

rehandalal:
r?

### DIFF
--- a/extension/content/less/cm-ndt.less
+++ b/extension/content/less/cm-ndt.less
@@ -44,6 +44,8 @@
     border: 1px solid var(--cm-border);
     border-radius: @input-border-radius;
     color: var(--cm-color);
+    font-family: @font-family-monospace;
+    font-variant-ligatures: no-contextual;
     line-height: 1.4375;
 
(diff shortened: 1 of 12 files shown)


extension/content/less/cm-ndt.less
```diff
@@ -44,6 +44,8 @@
     border: 1px solid var(--cm-border);
     border-radius: @input-border-radius;
     color: var(--cm-color);
+    font-family: @font-family-monospace;
+    font-variant-ligatures: no-contextual;
```
mythmon, line 48: Not a fan of ligatures?

mythmon (approved):
bors merge

extension/content/less/cm-ndt.less
rehandalal: I am but I think they should be optional with "off" being the default.

rehandalal: I'll be adding UI for extension settings that will let us customize some of the appearance stuff like this.

PR #159, opened 2020-05-03
Refresh auth tokens before they expire

### LINKED ISSUE #68: Handle refreshing auth tokens
In #51 we implement authentication but not handling refreshing the auth tokens. We will need to implement this.

rehandalal:
A few features:
- Checks if any tokens have expired every 10 minutes. If they are going to expire within 10 minutes of the check they are refreshed. If the refresh fails (typically because user interaction is required) the user is logged out.
- Login is now attempted non-interactively first so there's no pop up with you are already signed in with SSO. If action is required a pop up will be displayed.

Fixes #68 

r?

### DIFF
--- a/extension/content/contexts/environment.js
+++ b/extension/content/contexts/environment.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import PropTypes from "prop-types";
 import React from "react";
 import { Route, Redirect, Switch, useParams } from "react-router-dom";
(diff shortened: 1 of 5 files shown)


extension/content/contexts/environment.js
```diff
@@ -315,7 +353,9 @@ export async function login(dispatch, selectedKey, environment) {
     url: webAuth.client.buildAuthorizeUrl({
       state,
       nonce,
+      ...(details.interactive === false ? { prompt: "none" } : {}),
```
rehandalal: unsure about this pattern. it saves a bunch of lines of code but maybe it's not super clear to read?

alternative:
```js
const buildUrlOptions = {
  none,
  state,
}

if (details.interactive === false) {
  buildUrlOptions.prompt = "none";
}

webAuth.client.buildAuthorizedUrl(buildOptions);
```

mythmon: It's a neat trick, but I think the more verbose version is better here. For one, it doesn't really leave any room for commentary, and it would scale badly if we have lots of stuff that varies (like in the webpack config).

mythmon (approved):

extension/content/contexts/environment.js
rehandalal: as an aside... i did actually use this trick in that react devtools patch that you merged:
https://github.com/mozilla-extensions/normandy-devtools/commit/33e503ebde706ee419f03eed25fdaf9f370dda8f#diff-11e9f7f953edc64ba14b0cc350ae7b9dR48

should I undo that?

rehandalal:
waiting on https://github.com/mozilla-extensions/xpi-manifest/pull/37 before landing...

rehandalal:
bors r=@mythmon

extension/content/contexts/environment.js
mythmon: I don't think it's worth changing there.

PR #160, opened 2020-05-04
Add support for react-devtools

rehandalal:
This turned out to have a way simpler solution than expected. 

I built a whole reverse proxy setup and then realized all it was doing was importing a script that opened a socket connection. So I just bundled the connection script directly.

One caveat is that the connection script must be loaded before React DOM hence the sorting function.

r?


mythmon:
WFM

bors merge

PR #166, opened 2020-05-04
Add fix for versioning in __BUILD__ and update docs

### LINKED ISSUE #171: Can't access property `getRecipeSuitabilities`
This issue occurs on the details page and appears in the latest stage signed build:
https://github.com/mozilla-extensions/normandy-devtools/commit/44da8a146789ad645ef7688225509dda463f3813

image

rehandalal:
Fixes #171 

r?

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -75,3 +75,14 @@ Now simply reload the extension. It should connect to React Devtools.
 Releases are built automatically and available on the
 [releases pages on Github](https://github.com/mozilla/normandy-devtools/releases).
 There is no manual build process, since it relies on having secrets to sign the add-on.
+
+### Development releases
+
+We now provide automatically built and signed XPIs for development purposes. These can 
+be found by looking at the list of checks for the commit you want an XPI for and
+click through to the TaskCluster task for "dep-signing-normandy-devtools".
+
+You should now be able to find an XPI file in the list of artifacts.
+
+To be able to install these XPIs you need to go to `about:config` and add a preference
+named `xpinstall.signatures.dev-root` as a boolean set to `true`.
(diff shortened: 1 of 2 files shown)


mythmon:
Can you remind me how this will work for TC builds? Since last time this method broke because we couldn't take the merge commit before TC built it. Is that something Aki is working on?

Edit: specifically for tagged builds, like 2.0.0-alpha.1

rehandalal:
oh right... i forgot we made that change to deal with TC... put this on ice while i explore...

rehandalal:
This should work ~once https://github.com/mozilla-extensions/xpi-manifest/pull/38 lands~ now.

docs/dev.md
```diff
@@ -74,4 +74,15 @@ Now simply reload the extension. It should connect to React Devtools.
...
+### Development releases
+
+We now provide automatically built and signed XPIs for development purposes. These can 
+be found by looking at the list of checks for the commit you want an XPI for and
+click through to the TaskCluster task for "dep-signing-normandy-devtools".
+
+You should now be able to find an XPI file in the list of artifacts.
+
+To be able to install these XPIs you need to go to `about:config` and add a preference
+named `xpinstall.signatures.dev-root` as a boolean set to `true`.
```
rehandalal: All my AMO signed extensions got disabled this morning:
image

This is fixed by setting `xpinstall.signatures.required` to `false`. Setting this to false removes the need for the other setting. I'm pretty sure this is how I had my old profile set up, but is this a known issue?

mythmon: Yeah, this is the system working as intended. With `xpinstall.signatures.dev-root` set to true, Firefox uses an entirely different root cert to validate add-ons. It would be nice if it would accept both. I think we should explain the problem and advise not using testing builds on real profiles instead of suggesting we disable add-on signature checking.

mythmon (approved):

rehandalal:
I've updated the docs to reflect some lessons from #171

mythmon:
bors merge

PR #173, opened 2020-05-05
added cloning

### LINKED ISSUE #66: Add clone functionality
We should be able to clone an existing recipe to a new recipe.

tiftran:
Decided to put the button beside the edit recipe button, fixes #66 

Screen Shot 2020-05-05 at 4 46 22 PM


extension/content/components/recipes/details/DetailsHeader.js
```diff
@@ -127,6 +134,14 @@ export default function DetailsHeader() {
         {viewExperimentButton}
         {requestApprovalButton}
         {statusToggleButton}
+        <IconButton
+          appearance="primary"
+          className="ml-1"
+          icon={<Icon icon="copy" />}
+          onClick={handleCopyClick}
+        >
+          Copy Recipe
```
rehandalal: "Copy" is a bit confusing to me. Especially with that icon it's ambiguous whether you're copying the recipe to the clipboard or "making a copy". I think "Clone" or "Duplicate" are probably better choices for verbiage.

extension/content/components/recipes/details/DetailsHeader.js
```diff
@@ -127,6 +134,14 @@ export default function DetailsHeader() {
...
+        </IconButton>
```
rehandalal: The header is already getting pretty full in "half-width" view. I think a better solution for features like this that are less "everyday" use is to put a new button with an `ellipsis-h` icon at the end of the header. It should not be a primary button and clicking it should give you a dropdown with a submenu of tasks. Somewhat similar to the dropdown you get when you click on your avatar in the top right corner.

extension/content/components/recipes/details/DetailsHeader.js
```diff
@@ -26,6 +26,13 @@ export default function DetailsHeader() {
     history.push(`/${environmentKey}/recipes/${recipeId}/edit`);
   };
 
+  const handleCopyClick = () => {
+    history.push({
+      pathname: `/${environmentKey}/recipes/new`,
+      state: { data },
```
rehandalal: I actually think this should live at `/${environmentKey}/recipes/${recipeId}/clone` and then it can use the data right from the context.

rehandalal (changes requested):
a couple changes to the approach

extension/content/components/recipes/details/DetailsHeader.js
```diff
@@ -136,6 +142,21 @@ export default function DetailsHeader() {
...
+            <Popover>
+              <Button appearance="subtle" onClick={handleCopyClick}>
+                Clone Experiment
+              </Button>
+            </Popover>
+          }
+          trigger="click"
+        >
+          <div className="ml-1">
+            <Icon icon="ellipsis-h" />
```
rehandalal: I would make this:
```js
<IconButton icon={<Icon icon="ellipsis-h" />} />
```

extension/content/utils/normandyApi.js
```diff
@@ -74,10 +74,10 @@ export default class NormandyAPI extends API {
     });
   }
 
-  async saveRecipe(id, data) {
+  async saveRecipe(id, path, data) {
```
rehandalal: I don't think that `normandyApi` should concern itself with the `path`. We should deal with this in the form save method and pass a `null` id if we are cloning.

extension/content/components/recipes/form/RecipeFormHeader.js
```diff
@@ -31,11 +31,15 @@ export default function RecipeFormHeader() {
...
   const saveRecipe = (comment) => {
     const { action, comment: _omitComment, ...cleanedData } = data;
-    const requestSave = normandyApi.saveRecipe(recipeId, {
-      ...cleanedData,
-      comment,
-      action_id: action.id,
-    });
+    const requestSave = normandyApi.saveRecipe(
+      recipeId,
+      history.location.pathname,
```
rehandalal: once you get rid of `path` from `saveRecipe` this could maybe update to something like:
```js
const isCloning = history.location.pathname.match(/\/clone\/?$/);
const requestSave = normandyApi.saveRecipe(
  isCloning ? null : recipeId,
  {
    ...cleanedData,
```

extension/content/components/recipes/form/RecipeFormHeader.js
```diff
@@ -31,11 +31,15 @@ export default function RecipeFormHeader() {
 
   const saveRecipe = (comment) => {
     const { action, comment: _omitComment, ...cleanedData } = data;
```
rehandalal: as an aside I think we can actually get rid of `comment: _omitComment`. since we overwrite it below anyway, we do not need to clean it from the data.

rehandalal (changes requested):
just a few more little tweaks and i think we're good

extension/content/components/recipes/details/DetailsHeader.js
```diff
@@ -136,6 +142,21 @@ export default function DetailsHeader() {
...
+              <Button appearance="subtle" onClick={handleCopyClick}>
+                Clone Experiment
+              </Button>
+            </Popover>
+          }
+          trigger="click"
+        >
+          <div className="ml-1">
+            <IconButton icon={<Icon icon="ellipsis-h" />} />
+          </div>
```
rehandalal: nit: I don't think this `div` adds any value I'd move the `className` to the button.

rehandalal (approved):
last tiny nit. r+wc

bors d=@tiftran

PR #174, opened 2020-05-06
Track VPN status

rehandalal:
r?


extension/content/utils/fetchWithTimeout.js
```diff
@@ -0,0 +1,27 @@
+export default function fetchWithTimeout(url, options = {}, duration = 5000) {
```
rehandalal: the need for this appears to be to partially address the HTTP/2 TCP issue we discovered. if we can find a workaround for that we likely don't need (or want) a timeout on the fetch.

rehandalal:
Blocked on:
https://bugzilla.mozilla.org/show_bug.cgi?id=1635935

Depending on the outcome of that we may need to consider another approach. Or perhaps even in the meanwhile.

extension/content/utils/fetchWithTimeout.js
rehandalal: on second thought i think we should keep this

rehandalal:
this is ready for review. its not great until the mentioned bug is resolved but at least better than nothing.

extension/content/contexts/environment.js
```diff
@@ -276,10 +336,16 @@ export function useSelectedEnvironmentAuth() {
...
+export function useSelectedEnvironmentConnectionStatus() {
+  const { state } = React.useContext(environmentContext);
+  return state.connectionStatus[state.selectedKey];
+}
+
 export function useSelectedNormandyEnvironmentAPI() {
   const environment = useSelectedEnvironment();
   const auth = useSelectedEnvironmentAuth();
-  return new NormandyAPI(environment, auth);
+  const connectionStatus = useSelectedEnvironmentConnectionStatus();
```
mythmon: Why are these three separate things? Why not something like

```js
const {environment, auth, connectionStatus} = useSelectedEnvironmentInfo()
```

As the context grows it seems strange to keep adding these very specific selectors. Is there some advantage to it?

extension/content/utils/fetchWithTimeout.js
mythmon: Yeah. I think it would even be good if timeout was the default way we made requests everywhere.

extension/content/contexts/environment.js
```diff
@@ -196,7 +208,55 @@ export function EnvironmentProvider({ children }) {
...
+          status: false,
+          key,
+        });
+      });
+
+      if (status === "up") {
+        Object.entries(ENVIRONMENTS).forEach(async ([key, environment]) => {
+          let status = true;
+          try {
+            await fetchWithTimeout(
```
mythmon: It may also be worth checking that the body is at least vaguely what we expect. Maybe check if it is valid JSON?

mythmon (changes requested):

extension/content/utils/fetchWithTimeout.js
rehandalal: yeah, makes sense

extension/content/contexts/environment.js
rehandalal: yeah i think i had already done it so i was repeating the pattern but i was also starting to feel like there were too many of these

extension/content/contexts/environment.js
```diff
@@ -261,32 +318,42 @@ export function useEnvironments() {
...
+  return new NormandyAPI(environment, auth, connectionStatus);
 }
 
 export function useSelectedExperimenterEnvironmentAPI() {
-  const environment = useSelectedEnvironment();
+  const { environment } = useSelectedEnvironmentState();
   return new ExperimenterAPI(environment);
 }
 
+async function checkVPNStatus(normandyApi, maxAttempts, currentAttempt = 0) {
```
mythmon, line 342: This ended up being pretty clean, nice job.

mythmon (approved):
This works great. Toggling VPN on and off was noticed by NDT and switched reasonably quickly.

My one concern is that the red dot when you don't have VPN looks like there is a notification, and there isn't anything that explains what's going on. If it stays as a dot, I'd suggest adding a hover text, and some text inside the environment switcher.

Alternatively, since NDT works pretty well without VPN, we could make it have no indicator when VPN is off, and show something like a green lock when VPN is on? This would probably also mean making lots of the other parts of the UI VPN-aware, like the Edit Recipe button and/or page.

mythmon (approved):

mythmon:
bors merge

PR #176, opened 2020-05-08
added comment modal

### LINKED ISSUE #75: Require a comment when saving a revision
We are going to treat the comment field on revisions like a commit message for each revision.

To achieve this I think we should:
- Make sure the comment field is blank when editing/creating a new recipe because the comment is unique to each revision
- Pop up a modal when the user clicks the "Save" button prompting them to enter a comment.
- Revision does not save until the user enters a comment and hits a new save button on the modal.

tiftran:
Fixes #75

### DIFF
--- a/extension/content/components/recipes/form/RecipeFormHeader.js
+++ b/extension/content/components/recipes/form/RecipeFormHeader.js
@@ -1,14 +1,13 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { useHistory, useParams } from "react-router-dom";
-import { Alert, Button, Icon, IconButton, Modal } from "rsuite";
+import { Alert, Button, Icon, IconButton, Input, Modal } from "rsuite";
 
 import {
   useEnvironmentState,
   useSelectedNormandyEnvironmentAPI,
 } from "devtools/contexts/environment";
 import { useRecipeDetailsState } from "devtools/contexts/recipeDetails";
-import GenericField from "devtools/components/recipes/form/fields/GenericField";
 
 export default function RecipeFormHeader() {
   const { recipeId } = useParams();
@@ -30,10 +29,11 @@ export default function RecipeFormHeader() {
     }
   };
 
-  const saveRecipe = () => {
-    const { action } = data;
+  const saveRecipe = (comment) => {
+    const { action, comment: _omitComment, ...rest } = data;
     const requestSave = normandyApi.saveRecipe(recipeId, {
-      ...data,
+      ...rest,
+      comment,
       action_id: action.id,
     });
 
@@ -46,10 +46,6 @@ export default function RecipeFormHeader() {
         console.warn(err.message, err.data);
         Alert.error(`An Error Occurred: ${JSON.stringify(err.data)}`, 5000);
       });
-  };
-
-  const closeSaveModal = () => {
-    setShowCommentModal(false);
   };
 
   const handleBackClick = () => {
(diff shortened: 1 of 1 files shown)


mythmon:
If I'm reading this right, it only shows a comment for changes to an existing recipe? That doesn't seem right to me, since a large portion of recipes only ever have one revision.

tiftran:
@mythmon  lemme go fix that.... after rereading https://github.com/mozilla-extensions/normandy-devtools/issues/75, I feel like I probs interpreted it wrong

extension/content/components/recipes/form/RecipeFormHeader.js
```diff
@@ -71,6 +81,45 @@ export default function RecipeFormHeader() {
           Save
         </IconButton>
       </div>
+      <SaveModal
+        show={showCommentModal}
+        onClose={closeSaveModal}
```
rehandalal: this is a confusing name since it is not an event handler but rather a method that is actually closing the modal. perhaps we should just pass through `setShowCommentModal` as a property named `setShowModal` and then it can decide whether to set `true` or `false` in the modal component itself.

extension/content/components/recipes/form/RecipeFormHeader.js
```diff
@@ -71,6 +81,45 @@ export default function RecipeFormHeader() {
...
+    onClose();
+  };
+
+  return (
+    <Modal show={show} onHide={onClose}>
+      <Modal.Header>
+        <Modal.Title>Save Recipe</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <GenericField label="Comment on Revision" name="comment" />
```
rehandalal: A couple things about this field.

Using a generic field means its going to populate with the previous value which is not the behavior we want. This comment should be blank whenever the user is creating a new revision since it is like a commit message and unique to the revision.

Also I think it should be a textarea not a text input. 

Finally I think we should not have a label and instead have some placeholder text instead:
```js
<textarea placeholder="Please describe the changes you are making&hellip;" />
```

extension/content/components/recipes/form/RecipeFormHeader.js
```diff
@@ -1,47 +1,57 @@
...
-          console.warn(err.message, err.data);
-          Alert.error(`An Error Occurred: ${JSON.stringify(err.data)}`, 5000);
-        });
+      setShowCommentModal(true);
     } catch (err) {
       Alert.error(err.message);
     }
   };
 
+  const saveRecipe = () => {
```
rehandalal: we could potentially add `comment` as an argument on this function and then you don't need to manipulate the context/state and the modal would handle the state of that field itself.

rehandalal (changes requested):
nice work! this is generally the right direction but needs some tweaks.

extension/content/components/recipes/form/RecipeFormHeader.js
```diff
@@ -30,10 +29,11 @@ export default function RecipeFormHeader() {
     }
   };
 
-  const saveRecipe = () => {
-    const { action } = data;
+  const saveRecipe = (comment) => {
+    const { action, comment: _omitComment, ...rest } = data;
```
rehandalal: `rest` is a bit of a ambiguous variable name. i would go with something like `cleanedData`.

rehandalal (approved):
one nit, but r+wc

rehandalal:
bors delegate=@tiftran

PR #201, opened 2020-05-20
Add ESLint import plugin

rehandalal:
I finally figured out how to get eslint-plugin-import to address my OCD. 

r?


.therapist.yml
```diff
@@ -1,22 +1,23 @@
 actions:
   eslint:
     description: eslint
-    run: yarn eslint {files}
-    fix: yarn eslint --fix {files}
+    run: ./node_modules/.bin/eslint --color --max-warnings 0 {files}
```
mythmon, line 4: Why this set of changes?

extension/content/side-effects.js
```diff
@@ -0,0 +1,18 @@
...
+import yaml from "highlight.js/lib/languages/yaml";
+
+// Register highlight.js languages
+highlightjs.registerLanguage("javascript", javascript);
+highlightjs.registerLanguage("json", json);
+highlightjs.registerLanguage("yaml", yaml);
+
+// Mode for Code Mirror
+// eslint-disable-next-line no-unused-vars
+import { Controlled } from "react-codemirror2"; // Imported for side effect
```
mythmon: I don't think we need `{ Controlled }` here, or the eslint exception.

mythmon (changes requested):

.therapist.yml
rehandalal: - the switch from yarn to the node_modules path was because yarn adds all this extra noise to the therapist output. 
- `--color` is just to force color output which improves legibility (imo) when there's a long list of errors.
- `--max-warnings` could go in another PR but since I was making changes I figured I'd sneak this in. It addresses something I found in another branch where I was converting stuff to typescript where there were a bunch of ESLint warnings like `Missing return type on function` and `Unexpected any. Specify a different type`.

extension/content/side-effects.js
rehandalal: I should add a better comment there but it's required for the next couple lines of imports to actually work.

mythmon (approved):

mythmon:
bors merge

PR #211, opened 2020-05-26
recipe form tests

tiftran:


jest.config.js
```diff
@@ -10,7 +12,7 @@ module.exports = {
   setupFiles: ["<rootDir>/tests/conftests.js"],
   transform: {
     "^.+\\.js$": "babel-jest",
-    "^.+\\.tsx?": "ts-jest",
+    "^.+\\.tsx?": "babel-jest",
```
rehandalal: if we're using `babel-jest` for both we can just combine these two lines like:
```js
"^.+\\.[jt]sx?$": "babel-jest",
```

tests/factories/recipeFactory.js
```diff
@@ -0,0 +1,50 @@
+import faker from "faker";
+import { Factory, SubFactory, Field } from "./factory";
+import {
+  versionFilterObjectFactory,
+  channelFilterObjectFactory,
+} from "./filterObjectFactory";
+
+class RecipeFactory extends Factory {
```
rehandalal: It looks like you are creating a revision through this factory and not a recipe. As such this should be called `RevisionFactory`. Additionally I think it is important that these factories actually return a full revision object as we would expect the API to. This helps reduce the odds of some side effect caused by extra data goes unnoticed in the tests.

tests/factories/recipeFactory.js
```diff
@@ -0,0 +1,50 @@
...
+  getFields() {
+    return {
+      action: {},
```
rehandalal: you probably want an `ActionFactory` to generate this

tests/factories/recipeFactory.js
```diff
@@ -0,0 +1,50 @@
...
+
+  postGeneration() {
+    const versionFO = versionFilterObjectFactory.build();
+    const channelFO = channelFilterObjectFactory.build();
+    this.data.filter_object = [
+      ...this.data.filter_object,
+      versionFO,
+      channelFO,
+    ];
+  }
```
rehandalal: i don't think the factory should by default generate a bunch of filter objects and be populating them. i think if filter objects are required they should be explicitly added.

as an aside: if the user has explicitly provided a `filter_object` array in data then we should never extend that. in this pattern there is no way to explicitly set the filter objects to an empty array or have filter objects that don't include channel or version.

rehandalal: one thing you may use the `postGeneration` hook for is to generate action arguments based on the action selected in `this.data`. this would require you to create a bunch of action arguments factories for the different types of actions like `ConsoleLogArgumentsFactory`.

tests/factories/recipeFactory.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export class ConsoleLogRecipeFactory extends Factory {
```
rehandalal: We should have just one generic `RecipeFactory` and this one will likely need to be heavily reliant on `postGeneration`.

We'd want options for `actionName`, `actionArguments`, `filterObjects`, `isApproved`, `isEnabled`, etc that should all customize the build in `postGeneration`.

tests/factories/recipeFactory.js
```diff
@@ -0,0 +1,50 @@
...
+  postGeneration() {
+    const action = { name: "console-log", id: 1 };
+    const args = { message: faker.lorem.words() };
+
+    this.data.latest_revision = {
+      ...this.data.latest_revision,
+      arguments: args,
+      action,
+    };
+  }
```
rehandalal: similar to the above `postGeneration`, this pattern causes us to lose explicit control over the data. if overrides are required they should be explicitly requested through `options` which are the second parameter passed to the build function. something like:
```js
RecipeFactory.build({}, { actionName: "console-log" });

class RecipeFactory {
  ...

  postGeneration() {
    if (this.options.actionName) {
      this.data.latest_revision = {
        ...this.data.latest_revision,
        action: ActionFactory.build({ name: actionName }),
      }
    }
  }
}
```

tests/factories/filterObjectFactory.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  postGeneration() {
+    const versionSet = new Set();
+    const setTotal = faker.random.number({ min: 1, max: 10 });
+    for (let i = 0; i < setTotal; i++) {
+      versionSet.add(faker.random.number({ min: 40, max: 100 }));
+    }
+
+    this.data.versions = [...this.data.versions, ...Array.from(versionSet)];
```
rehandalal, line 20: similar concern here about not being able to explicitly set versions. you could simply hard code a list as the default or add an option for `generateVersionsCount` which would be used instead of `setTotal`.

tests/factories/filterFactory.js
```diff
@@ -0,0 +1,87 @@
...
+  getFields() {
+    return {
+      channels: Channels,
+      countries: [],
+      locales: [],
+    };
+  }
+
+  postGeneration() {
+    const { countries, locales } = this.options;
```
rehandalal: i think these should be `generateCountriesCount`, `generateLocalesCount` and should have a default set.

rehandalal (changes requested):
nice work on this! i think there are a few things to work out with the factories but the tests are looking good! thanks for your work on this!

rehandalal:
also this needs master to be merged in

tests/factories/filterObjectFactory.js
```diff
@@ -0,0 +1,121 @@
...
+      versions: [],
+    };
+  }
+
+  postGeneration() {
+    const versionSet = new Set();
+    const { generateVersionsCount } = this.options;
+    for (let i = 0; i < generateVersionsCount; i++) {
+      versionSet.add(faker.random.number({ min: 40, max: 100 }));
+    }
```
rehandalal: nit: its possible if you set `generateVersionsCount` to `n` that you will actually get `< n` values due to de-duplication. a better approach is probably to use a `while`:
```js
const { generateVersionsCount = 0 } = this.options;
while (versionSet.size < generateVersionsCount) {
  versionSet.add(fake.random.number({ min: 40, max: 100 });
}
```

rehandalal (approved):
one small nit and this is good to go.

thanks for all your hard work on this!! r+wc

bors d=@tiftran

PR #215, opened 2020-05-28
added generic FO textbox

tiftran:


mythmon:
The JSON editor sets it's value to `null` when there is invalid JSON, which confused me while testing this. I included an extra comma in the field and it wasn't included at all in the request sent to the server.

2020-05-28-090114_screenshot

This has been ok with other fields because they are required and so the server catches the error, but this field would be easy to mess up. We should at least provide some visual feedback that the field has errors. Ideally it would prevent submission entirely while there are syntax errors.

That being said, I was able to make a recipe with a custom filter object. It works! (ext+normandy://dev/recipes/134) It's not the friendliest thing, but that's kind of the whole point. Nice job!

tiftran:
@mythmon I added a badge for invalid json, was wondering if that's what you had in mind
Screen Shot 2020-05-28 at 9 34 45 AM

mythmon:
Looks great!

extension/content/components/recipes/form/filters/partitionFO.js
```diff
@@ -0,0 +1,21 @@
...
+    ([knownFO, additionalFO], fo) => {
+      if (fo && KNOWN_FILTER_TYPES.includes(fo.type)) {
+        return [[...knownFO, fo], additionalFO];
+      }
+
+      return [knownFO, [...additionalFO, fo]];
+    },
+    [[], []],
+  );
+};
```
rehandalal: I'm not sure why we need this function? Its use outside `FallbackFO` seem unnecessary?

rehandalal: also if we do need it, this is not a component and is more in the shape of a utility function and so it should not live under `components` unless maybe it's in a single file that uses it like `FallbackFO.js` as a helper function that is not exported.

rehandalal: if we do decide to keep this function I think it should return an object roughly in the shape of:
```js
{
  known: [],
  unknown: [],
}
```

or something to that effect rather than an array as that's less opaque

extension/content/components/recipes/form/filters/SamplingOptions.js
```diff
@@ -249,7 +250,8 @@ function getFilterObjectFromData() {
 
   let filterObject;
   if (data.filter_object) {
-    filterObject = data.filter_object.find((fo) =>
+    const [filterObjects, _omitAdditionalFO] = partitionFO(data.filter_object);
```
rehandalal: fyi: you dont need the `_omitAdditionalFO`. since you are not trying to access an item in the array with a higher index you can simply omit it like:
```js
const [filterObjects] = partitionFO(data.filter_object);
```

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,59 @@
...
+  const inValidBadge = () => {
+    if (invalidJSON) {
+      return <Badge content="Invalid JSON" />;
+    }
+
+    return null;
+  };
+
+  let key = data.recipe ? data.recipe.id : "create";
+  key += "FO";
```
rehandalal: what do we gain by adding this to the end of `key`?

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,59 @@
...
+          filter_object: newFilterObjects,
+        },
+      });
+      setInvalidJson(false);
+    } else {
+      setInvalidJson(true);
+    }
+  };
+
+  const inValidBadge = () => {
```
rehandalal: nit: `inValid` is strange camelcasing when the word is `invalid` and not `in valid`.

rehandalal: also i dont think this needs to be a function? it could simply be written as:
```js
let invalidBadge = null;
if (invalidJSON) {
  invalidBadge = <Badge content="Invalid JSON" />;
}
```

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    return null;
+  };
+
+  let key = data.recipe ? data.recipe.id : "create";
+  key += "FO";
+  return (
+    <FormGroup>
+      <ControlLabel>Additional Filter Objects {inValidBadge()}</ControlLabel>
```
rehandalal: if `invalidBadge` is not a function this needs to be updated

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,59 @@
...
+export default function FallbackFO() {
+  const data = useRecipeDetailsData();
+  const dispatch = useRecipeDetailsDispatch();
+  const [invalidJSON, setInvalidJson] = React.useState(false);
+
+  let knownFO = [];
+  let additionalFO = [];
+  if (data.filter_object) {
+    [knownFO, additionalFO] = partitionFO(data.filter_object);
+  }
```
rehandalal: i think if you just do the split here you never need to call `partitionFO` ever again so I'm not sure that even needs to be a helper function. you can simply reuse `knownFO` from this scope in `handleChange`.

rehandalal (changes requested):

extension/content/components/recipes/form/filters/partitionFO.js
tiftran: oooh you're right it isnt needed, this lingered from my first attempt where i tried to save the json blob regardless of whether it was valid or not, so FO was an array of objects and strings and it got weird... didn't go back to check if it was still needed, my bads

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,72 @@
...
+      ([knownFO, additionalFO], fo) => {
+        if (fo && KNOWN_FILTER_TYPES.includes(fo.type)) {
+          return [[...knownFO, fo], additionalFO];
+        }
+
+        return [knownFO, [...additionalFO, fo]];
+      },
+      [[], []],
+    );
+  }
```
rehandalal: i think as it stands this is a little hard to read and could be simplified to something like:
```js
const knownFO = [];
const additionalFO = [];
if (data.filter_object) {
  data.filter_object.forEach((fo) => {
    if (KNOWN_FILTER_TYPES.includes(fo.type)) {
      knownFO.push(fo);
	} else {
      additionalFO.push(fo);
    }
  });
}
```

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,72 @@
...
+        type: ACTION_UPDATE_DATA,
+        data: {
+          ...data,
+          filter_object: newFilterObjects,
+        },
+      });
+      setInvalidJson(false);
+    } else {
+      setInvalidJson(true);
+    }
```
rehandalal: i *think* this would result in a blank field reporting "invalid JSON" and that's not correct. it sounds like what you want to do here is update `JsonEditor` so that it's `onChange` event fires the callback with `(value, error)` so that it reports if a JSON parsing error actually occured and you can respond accordingly.

rehandalal (changes requested):

extension/content/components/recipes/form/filters/FallbackFO.js
tiftran: oooh you're right, will fix this

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -0,0 +1,83 @@
...
+  };
```
rehandalal: i think there's room to refactor this to be more DRY. also there is the issue of what if `value` is not an array. i think this fits into the broader issue of surfacing form errors better and we should probably have a discussion about this.

Just to address the DRY bits:
```js
const handleChange = (value, err) => {
  if (err) {
    setInvalidJson(true);
  } else {
    setInvalidJson(false);
    let newFilterObjects = [...knownFo];
    if (value) {
      newFilterObjects = newFilterObjects.concat(value);
    }
    dispatch({
      type: ACTION_UPDATE_DATA,
      data: {
        ...data,
        filter_object: newFilterObjects,
      },
    });
  }
}
```

rehandalal (changes requested):

extension/content/components/recipes/form/filters/FallbackFO.js
tiftran: Made the suggested changes. When you say surfacing form errors, not sure if you meant the fallback editor specifically or all the other forms as well or whether changes should be done within this PR?

rehandalal: I don't think we should handle all the form errors in this PR but I think it's prudent to come up with a plan for dealing with all form errors (I think there are two classes of errors: server-side and client-side validation errors) that we need to deal with and we should address this client-side error with a broader framework in mind.

extension/content/contexts/recipeDetails.js
```diff
@@ -84,3 +92,8 @@ export function useRecipeDetailsImportInstructions() {
   const { state } = React.useContext(recipeDetailsContext);
   return state.importInstructions;
 }
+
+export function useRecipeDetailsClientErrors() {
```
rehandalal: I think what might be better is to have this be `useRecipeDetailsErrors` and have it return `{ clientErrors }` for now, eventually it will be an object with both `clientErrors` and `serverErrors`.

extension/content/contexts/recipeDetails.js
```diff
@@ -32,6 +34,12 @@ function reducer(state, action) {
         importInstructions: action.importInstructions,
       };
 
+    case ACTION_UPDATE_CLIENT_ERRORS:
```
rehandalal, line 38: I think it might be better to have two actions and have them both take an action name:
```js
case ACTION_UPDATE_CLIENT_ERRORS:
  return {
    ...state,
    clientErrors: {
      ...state.clientErrors,
      [action.name]: [action.errors],
    },
  }

case ACTION_REMOVE_CLIENT_ERRORS:
  const { [action.name]: _omit, ...clientErrors } = state.clientErrors;
  return {
    ...state,
    clientErrors,
  }
```

rehandalal (changes requested):

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -22,9 +23,10 @@ const KNOWN_FILTER_TYPES = [
 
 export default function FallbackFO() {
   const data = useRecipeDetailsData();
-  const errors = useRecipeDetailsClientErrors();
+  const errors = useRecipeDetailsErrors();
+  const { clientErrors } = errors;
```
rehandalal: this could be condensed to:
```js
const { clientErrors } = useRecipeDetailsErrors();
```

extension/content/components/recipes/form/filters/FallbackFO.js
```diff
@@ -42,6 +44,8 @@ export default function FallbackFO() {
   const notAnArray = "Filter Object(s) is not contained in an array";
 
   const validateFO = (value, err) => {
+    let action = ACTION_REMOVE_CLIENT_ERRORS;
+    let actionArgs = { name: "filter_object" };
```
rehandalal: i think these should move down below to where the `if (foErrors) {` line is (line 60).

rehandalal (changes requested):
this lgtm. just one other change i'd like to see here. you should not be able to submit the form if there are any client errors.

rehandalal (approved):
looks good! thanks!!

bors r+

PR #237, opened 2020-06-04
changed isHighVolume -> isHighPopulation

tiftran:
@mythmon pointed out that in devtools we have the field isHighVolume for pref experiments rather than the isHighPopulation, I think these changes should correct that.


mythmon (approved):
bors merge

PR #250, opened 2020-06-09
removed unused lines in recipe list

tiftran:
this came from when "view recipe" used to show a modal instead of the recipe details page


rehandalal (approved):
bors r+

PR #253, opened 2020-06-09
change if cond for remove fo errors

tiftran:
I think this solves it

Regardless of whether the foerrors array is empty, it still seems to have changed to the update action rather than  keep the remove?

### DIFF
--- a/tests/RecipeForm.test.js
+++ b/tests/RecipeForm.test.js
@@ -40,6 +40,7 @@ describe("The `RecipeForm` component", () => {
     const versionForm = findForm(formGroups, "Version");
     const countriesForm = findForm(formGroups, "Countries");
     const localesForm = findForm(formGroups, "Locales");
+    const fallbackFOForm = findForm(formGroups, "Additional Filter Objects");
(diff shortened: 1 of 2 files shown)


rehandalal (changes requested):
whoops! let's add a test too.

tests/RecipeForm.test.js
```diff
@@ -358,4 +360,84 @@ describe("The `RecipeForm` component", () => {
...
+      getByText("Filter Object(s) is not contained in an array"),
+    ).toBeInTheDocument();
+    expect(saveButton).toHaveAttribute("disabled");
+
+    //ctrl z... the paste
+    foCodeBlock.dispatchEvent(new Event("focus"));
+    foCodeBlock.dispatchEvent(
+      new KeyboardEvent("keydown", { keyCode: 90, ctrlKey: true }),
+    );
+
```
tiftran, line 409: A kind of a round about way to change the value for generic FO....
I wasn't able to use`fireEvent`to change it, nor was I able to use an Input Event to change it either 😬

rehandalal (approved):
bors r+

PR #254, opened 2020-06-11
enforce isEnrollmentPaused for required recipes

tiftran:


rehandalal (changes requested):
im not sure this is the right solution. i think the only time this is an issue is when importing from experimenter and there is no `isEnrollmentPaused` key passed through in the import data. i think this is easily remedied by making sure the initial action arguments have `isEnrollmentPaused` in `extension/content/components/recipes/form/ActionArguments.js` (which you did in this PR) and then updating the import data appropriately at import time:

---

`extension/content/components/pages/RecipeFormPage.js`: (line 31)
```js
    } else if (experimenterSlug) {
      normandyApi.fetchAllActions().then((actions) => {
        experimenterApi
          .fetchRecipe(experimenterSlug)
          .then(({ comment, action_name, ...recipeData }) => {
            setData({
              ...recipeData,
              arguments: {
                ...INITIAL_ACTION_ARGUMENTS[action_name],
                ...recipeData.arguments,
              },
              action: actions.find((a) => a.name === action_name),
            });
            setImportInstructions(comment);
          });
      });
```

rehandalal (approved):
thanks!!!

bors r+

PR #266, opened 2020-06-18
Calculate a testing client ID for bucket samples filters

mythmon:
I wanted to implement this in a place where I could re-use use instead of in Firefox's console.

It ends up looking like this:

image


rehandalal (dismissed):
very cool!

needs tests though.

mythmon:
Finally got tests that pass for this, and have high enough coverage. That was a bit tougher than I expected.

@tiftran Can you review this now, since Rehan is out?

tiftran (approved):
Looks good to me, and thank you for correcting my typo :)

PR #283, opened 2020-06-30
Ensure the writeable URL is used to read details when VPN is connected

rehandalal:
Fixes an issue where we don't actually use the writeable URL for VPN connected users

r?


tiftran (approved):
looks good to me

PR #295, opened 2020-07-06
Update babel requirements to only recent Firefox and Node

mythmon:
This means the code is built with fewer transforms, and so it is closer to the original code, making it easier to debug. It's also smaller and builds faster.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,15 +7,6 @@
   ],
   "plugins": [
     ["@babel/plugin-proposal-decorators", { "legacy": true, "loose": true }],
-    "@babel/plugin-proposal-object-rest-spread",
-    "@babel/plugin-proposal-async-generator-functions",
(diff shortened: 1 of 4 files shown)


mythmon:
My changes confusingly changed test coverage, even though I didn't touch the code in question. I added an unrelated test to offset the difference. This is definitely goofy, and something probably shouldn't do?

tiftran (approved):

tiftran:
oooh i didn't see the tests that you add... hooray for more tests?
I don't think that reducing coverage blocks from merging.... so in cases where there isn't an actual implementation change, I think it makes sense that there aren't any tests?

mythmon:
Ah, I assumed that the failed CI status from Codecov would block things, but I guess bors doesn't actually care about that. That's good to keep in mind in the future, but yeah, I guess yay more tests.

PR #315, opened 2020-07-22
Publish docs explicitly to gh-pages

mythmon:
Publishing the docs/ folder automatically only works if your default branch is named master.

Fixes #315


rehandalal (approved):
good catch!

bors r+

PR #330, opened 2020-07-29
Fix bug in direct bucket sample method

### LINKED ISSUE #328: Client test id is wrong
The client test ID reported for bucket samples use the wrong inputs to brute force the filter, and so the wrong ID is reported.

mythmon:
The inputs of sampling arguments are JEXL expressions, so to provide a fake ID
we should pass it as a string. Additionally, it should be processed as a JEXL
expressoin by the `bucketSample` bypass method. Parsing it as JSON is close
enough for our purposes.

Fixes #328.

### DIFF
--- a/extension/content/utils/recipes.ts
+++ b/extension/content/utils/recipes.ts
@@ -48,11 +48,12 @@ export async function bruteForceBucketSample(
   const maxTrials = (filter.total / filter.count) * 10;
   for (let i = 0; i < maxTrials; i++) {
     const fakeClientId = `test-userId-${i}`;
+    const fakeClientIdStr = `"${fakeClientId}"`;
     const newInput = filter.input.map((inp) =>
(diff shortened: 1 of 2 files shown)


mythmon:
I'm not really sure how to write a test for this. I'm open to ideas but I think we should just merge this as is.

tiftran (commented):
Looks good to me? 
Not sure if I have the correct context,  but are the testing clientIds used to be able to be in one of the recipe buckets?

mythmon:
> Not sure if I have the correct context, but are the testing clientIds used to be able to be in one of the recipe buckets?

Yeah, this gives you a clientId that is guaranteed to be chosen by the bucket filter, which makes QA and testing easier.

tiftran (approved):

mythmon:
bors r= @tiftran

PR #359, opened 2020-08-17
Turn off dependabot auto-rebasing

mythmon:
Dependabot is rebasing PRs that don't have any conflicts, which makes bors only able to merge one at t ime. This makes managing Dependabot PRs in this repo a lot harder. Turning it off will mean we'll have to manually trigger Dependabot rebasing when it is needed, but that's pretty uncommon for us.


jaredlockhart (approved):
Yep looks right to me 👍

PR #363, opened 2020-08-20
Reorganize tests to integrate with Typescript better

mythmon:
* Moves tests into the extension/content directory, since they are all for that.
* Remove the specific tsconfig.json for tests as it is no longer needed.
* Fix or ignore many TypeScript problems that existed in tests because they were not being checcked.
* Update imports to use absolute project imports more often.


rehandalal (approved):
bors r+

PR #400, opened 2020-09-09
Move dependabot config to the correct location

mythmon:
After reviewing the docs further, I realized our dependabot config file is in the wrong place. This should fix the problem.


mythmon:
bor r+

tiftran (approved):

PR #417, opened 2020-09-15
Fix #404: in-page address bar is now editable

### LINKED ISSUE #404: In-page address bar is not editable
The in-page address bar that lists `ext+normandy://` addresses should be editable. A user should be able to type into it and then press enter to navigate to that desired page, just as in the normal address bar.

leplatrem:
Fix #404

With this change, the in-page addressbar is not constantly overwritten. The effect/hook was marked as depending from `extensionURL`, which becomes different than edited value after first stroke.


rehandalal (approved):
bors r+

PR #438, opened 2020-09-21
Switch to using the web+normandy:// protocol handler

### LINKED ISSUE #412: Switch to using web+normandy:// custom protocol handler
NDT currently uses `ext+normandy://` as a custom protocol handler for shortened links. With the new web version of NDT #411, we won't be able to use `ext+...` protocols. We can however register `web+...` protocol handlers.

Because of this, we should switch NDT to use a `web+...` style protocol handler, so the two systems can interoperate, regardless of which version someone uses.

mythmon:
Fixes #412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next
 
 - Add a one-click pause and request review button to recipe detail pages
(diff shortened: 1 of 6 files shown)


rehandalal (approved):
bors r+

mythmon:
bors r= @rehandalal

PR #439, opened 2020-09-23
Add Experimenter info to details page (fixes #407)

### LINKED ISSUE #407: Add Experimenter data to recipe detail page
If a recipe has a related Experimenter ticket, relevant data from that ticket should be included on the detail page for the recipe. Examples of useful data to bring in could include the internal description of the experiment, and the schedule details for experiments and rollouts.

leplatrem:
Fixes #407

* [x] Early feedback on strategy
* [x] Integrate Experimenter API client with environment switching
* [x] Determine fields and info to display
* [ ] Find out solution for UX and async load: «scrolling/flickering» issue (info panel first empty, then load, then scroll)
* [x] Pull branch names
* [x] Add tests


leplatrem:
@rehandalal @mythmon Do you think that this is roughly the right approach? 

I was wondering whether I should blend experimenter and normandy info at the `RecipeDetailsProvider` level. I thought that keeping a clear mapping of local types and values returned by the API would be less confusing. But there is at least one advantage to blend them: the page is loaded in a single step. 

What do you think? I'm happy to clean-up/change everything at this stage :)

mythmon:
I think this looks about right. I think that keeping the data sources separate is the right choice. The data is loaded separately, and trying to conflate it is just hiding that. Preventing the UI from jumping around and being in an inconsistent state is a UI concern. Maybe Rsuite's [Placeholder](https://rsuitejs.com/components/placeholder/) could help fill the space while things are loading?

One thing I'd point out is that not all recipes have Experimenter slugs, so this data shouldn't always be visible. It would probably be good to check for that before making requests to `/api/v1/experiments/undefined`.

One small thing I'd request: please consider using Typescript for new files, if it won't be too much of a hassle. Typescript has provided a lot of benefits to me on this project, and I'd like to see more code be well typed. You can simply name the file `foo.ts` or `foo.tsx` if you want to use JSX syntax in it. All of the import and export machinery will smoothly translate back and forth with no other changes.

leplatrem:
Screenshot from 2020-10-05 17-06-48
:man_shrugging:

mythmon:
I think I know why the test failed. They pass as well on my machine. I noticed that the failure that CI is getting is

```
 Expected: "Tue, 08 Sep 2020 00:00:00 GMT → Wed, 06 Jan 2021 01:00:00 GMT (120 days)"
 Received: "Tue, 08 Sep 2020 00:00:00 GMT → Wed, 06 Jan 2021 00:00:00 GMT (120 days)"
```

Being exactly one hour different made me think DST and timezones. I was also curious why the tests expected now + 120 days to advanced the hour by one as well. I dug around a bit, and eventually did this test:

```node
> d = new Date()
2020-10-06T18:48:45.349Z
// hour is 18
> d.setDate(d.getDate() + 120)
> d
2021-02-03T19:48:45.349Z
// hour is 19, as expected by current code

> d = new Date()
2020-10-06T18:49:17.597Z
// hour 18 again
> d.setUTCDate(d.getUTCDate() + 120)
> d
2021-02-03T18:49:17.597Z
> // this time it didn't change
```
I have a theory as to why this happened. The test is adding 120 local days to a **local** timezone. On our developer machines, that means a timezone that includes DST, and there is a DST transition going one hour forward in the next 120 days (November 1st in the US). CircleCI runs tests with the local time set to UTC, which does not observe DST, and so there is no "extra" hour.

Changing to `setUTCDate` and `getUTCDate` fixes this problem. However, I have two other suggestions:

1. Experimenter never deals in anything more granular than a day. We probably shouldn't be showing the hours or timezone for planned dates anyways.
2. Time is hard. Perhaps we should pull in a library like [Luxon](https://moment.github.io/luxon/) or [date-fns](https://date-fns.org/) to do the heavy lifting for us, and get better formatting options as a bonus. (Note that the moment.js docs now include this line: "We now generally consider Moment to be a legacy project in maintenance mode.". I don't think we should use it)

leplatrem:
I think it's good enough for a first iteration. If leveraging a lib is a blocker that'd be tomorrow :)

content/components/pages/RecipeDetailsPage.tsx
```diff
@@ -3,31 +3,67 @@ import { useParams } from "react-router-dom";
...
 import { RecipeDetailsProvider } from "devtools/contexts/recipeDetails";
 
 // export default
 const RecipeDetailsPage: React.FC = () => {
   const { recipeId } = useParams<{ recipeId: string }>();
   const normandyApi = useSelectedNormandyEnvironmentAPI();
-  const [data, setData] = React.useState({});
+  const experimenterApi = useSelectedExperimenterEnvironmentAPI();
+  const [recipeData, setRecipeData] = React.useState({
+    experimenter_slug: "",
```
mythmon: The experimenter slug is not always a string, as sometimes it is `null`. I don't think that modifying the default state to include an empty string is right here.

content/components/recipes/details/ExperimenterDetails.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <div className="mt-4">
+          <strong>Actual Schedule</strong>
+          <p>
+            {startDate ? startDate.toDateString() : "N/A"}
+            {endDate ? " → " + endDate.toDateString() : ""}
+          </p>
+        </div>
+        <div className="mt-4">
+          <strong>Branches</strong>
+          {variants.length ? (
```
mythmon: I don't think it is useful to list the variants here, as it spreads associated data across multiple places. The variant names are more useful when combined with the rest of the data associated with each. We have that info already in the "Branches" section of the relevant action arguments (Experimenter's variants are Normandy's branches). I think that instead we should try and associate these variant names with the branches where they already are.

content/contexts/experimenterDetails.tsx
```diff
@@ -0,0 +1,77 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+interface ExperimenterData {
```
mythmon, line 3: It *might* be worth it to put these types in `content/types/experimenter.ts`, if they get re-used elsewhere. Though considering they aren't currently exported, maybe that's not a concern.

content/contexts/experimenterDetails.tsx
```diff
@@ -0,0 +1,77 @@
...
+      type: ACTION_UPDATE_DATA,
+      data,
+    });
+  }, [data]);
+
+  // @ts-ignore
+  return <Provider value={state}>{children}</Provider>;
+}
+
+ExperimenterDetailsProvider.propTypes = {
```
mythmon: We probably shouldn't be mixing proptypes and Typescript.

content/tests/components/recipes/details/RecipeDetails.test.js
```diff
@@ -35,6 +36,26 @@ describe("The `RecipeDetailForm` component", () => {
     jest
       .spyOn(NormandyAPI.prototype, "fetchRecipe")
       .mockImplementation(() => Promise.resolve(recipe));
+
+    jest
+      .spyOn(ExperimenterAPI.prototype, "fetchExperiment")
+      .mockImplementation(() =>
+        Promise.resolve({
+          public_description: "Try this new thing",
```
mythmon: It may be worthwhile to make this a factory, in `content/tests/factories/experimenter.ts`.

content/tests/components/recipes/details/RecipeDetails.test.js
```diff
@@ -141,6 +162,30 @@ describe("The `RecipeDetailForm` component", () => {
...
+    const { getByText, findByText, findByTestId } = await render(<App />);
+    fireEvent.click(getByText("Recipes"));
+    fireEvent.click(await findByText(recipeData.latest_revision.name));
+
+    await waitFor(() =>
+      expect(ExperimenterAPI.prototype.fetchExperiment).toReturn(),
+    );
+    await waitFor(() => expect(NormandyAPI.prototype.fetchRecipe).toReturn());
+
+    expect(getByText("Experimenter Details")).toBeInTheDocument();
```
mythmon, line 178: I find it hard to be sure that this test is right since the experimenter definitions are pretty far away. Just reading this test it isn't clear that the proposed start date should be "Sep 08", for example.

I think that following the example of how recipe data is handled with `setup` might be easier: Then the experimenter definitions that matter can be defined in this test with a factory, while still abstracting out all the API mocking.

mythmon (changes requested):
I don't think using a library is a blocker here. I suggested it as a way to make things easier if you'd like, but doing it without the library is fine too.

content/contexts/experimenterDetails.tsx
leplatrem: I think I'll need some guidance, I was looking at `normandy-devtools/content/contexts/recipeDetails.js ` and https://github.com/mozilla-extensions/normandy-devtools/blob/c8d75d4650216e451a23fe176c1fff27735bcdb5/content/contexts/environment.tsx#L256-L258

mythmon: Oops, I guess I missed that in my conversion. The pattern I've using has been to leverage the `React.FunctionalComponent` (or `React.FC` for short) which can be parameterized on the type of its props. So in this case it would be something like

```ts
const ExperimenterDetailsProvider: React.FC<ExperimenterState> = ({ children, data }) => {
	// ...
}
```

`children` is automatically added to the type of the props by `React.FC`. This results in bad props being a build time problem instead of a quiet runtime problem.

content/components/recipes/details/ExperimenterDetails.tsx
leplatrem: Can we spin off an issue to follow-up on that? 

It looks like I will have to go through every kind of action and merge details there. I think this PR is big enough and would feel better to first merge this and go on with smaller stuff...

mythmon: Yeah, lets spin that off. I'm looking forward to getting this merged too.

content/contexts/experimenterDetails.tsx
```diff
@@ -0,0 +1,67 @@
...
+  variants: Record<string, string>;
+}
+
+interface ExperimenterState {
+  data: ExperimenterData;
+}
+
+interface ExperimenterState {
+  data: ExperimenterData;
+}
```
mythmon: These two definitions are duplicates, right?

content/components/recipes/details/ExperimenterDetails.tsx
```diff
@@ -0,0 +1,61 @@
...
+const ExperimenterDetails: React.FunctionComponent = () => {
+  const data = useExperimenterDetailsData();
+  return (
+    <CollapsibleSection
+      headerButtons={<></>}
+      title={
+        <>
+          <h6>Experimenter Details</h6>
+        </>
+      }
```
mythmon: The fragment (`<></>`) here isn't needed. That is only needed if you need a JSX element that has `n != 1` children.

content/contexts/experimenterDetails.tsx
```diff
@@ -0,0 +1,67 @@
...
+          ...action.data,
+        },
+      };
+
+    default:
+      return state;
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
mythmon: I *expect* that now that you have typed this as a `React.FC`, you don't need this ESLint exception anymore.

content/contexts/experimenterDetails.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const ExperimenterDetailsProvider: React.FC<ExperimenterState> = ({
+  children,
+  data,
+}) => {
+  /** @type {[React.ReducerState<any>, React.Dispatch<React.ReducerAction<any>>]} */
+  const [state, dispatch] = React.useReducer(reducer, initialState);
```
mythmon: I think you can avoid `any` here with

```suggestion
  const [state, dispatch] = React.useReducer<Reducer<ExperimenterState, ExperimenterAction>>(reducer, initialState);
```

Though then you'd need to define `ExperimenterAction`. I did something like this in `environment.tsx`. If this is tricky, feel free to leave it as is.

content/tests/components/recipes/details/RecipeDetails.test.js
```diff
@@ -89,6 +91,16 @@ describe("The `RecipeDetailForm` component", () => {
     return recipe;
   };
 
+  const experimenterSetup = (experiment) => {
+    if (!experiment) {
+      experiment = experimenterResponseFactory.build();
+    }
```
mythmon: It's a little more verbose, but since JS parameter defaults are evaluated at calltime, you can do

```suggestion
  const experimenterSetup = (experiment = experimenterResponseFactory.build()) => {
```

content/tests/factories/experiments.ts
```diff
@@ -0,0 +1,21 @@
...
+  { generateVariantsCount: number }
+>({
+  public_description: faker.lorem.sentence(),
+  proposed_start_date: faker.date.past().getTime(),
+  proposed_duration: faker.random.number(),
+  start_date: faker.date.past().getTime(),
+  end_date: faker.date.recent().getTime(),
+  variants: ({ generateVariantsCount = 1 }) =>
+    // eslint-disable-next-line prefer-spread
+    Array.apply(null, { length: generateVariantsCount }).map(() => ({
```
mythmon, line 17: I haven't seen this trick before

mythmon (approved):
There a couple things I think could be better, but this should work as is. The comments here are suggestions, feel free to implement which ever of them you'd like.

bors d+

PR #440, opened 2020-09-23
Don't calculate coverage for Dependabot PRs

mythmon:
Dependabot PRs often change test coverage even though there isn't any actual
difference. This makes the automated CI checks less useful since about half
of Dependabot PRs have a false-negative red X on the PR summary.


rehandalal (approved):
bors r+

mythmon:
Oops, this didn't actually work. Due to wrong bash syntax, it just stopped collecting coverage entirely. see #441.

PR #449, opened 2020-09-25
Hide pages that rely on extension powers in web-mode

mythmon:
- Enable code splitting and minification for the web build
- Hide extension-only pages in web env


content/.eslintrc.js
```diff
@@ -2,7 +2,7 @@
 module.exports = {
   globals: {
     __BUILD__: false,
-    ___ENV__: false,
+    __ENV__: false,
```
rehandalal, line 5: gah! nice catch

content/components/App.js
```diff
@@ -10,16 +10,31 @@ import {
...
+    ? null
+    : React.lazy(() => import("devtools/components/pages/FiltersPage"));
+const PrefStudiesPage =
+  __ENV__ === "web"
+    ? null
+    : React.lazy(() => import("devtools/components/pages/PrefStudiesPage"));
+const AddonStudiesPage =
+  __ENV__ === "web"
+    ? null
+    : React.lazy(() => import("devtools/components/pages/AddonStudiesPage"));
```
rehandalal, line 36: TIL: `React.lazy`. Neat!

webpack.config.js
```diff
@@ -186,6 +180,15 @@ function makeExtensionConfig(development) {
...
       /* indent width */ 2,
     ),
+
+    // Code splitting doesn't do any good in extensions
+    new webpack.optimize.LimitChunkCountPlugin({
+      maxChunks: 1,
+    }),
+
+    new webpack.DefinePlugin({
+      __ENV__: JSON.stringify("extension"),
```
rehandalal, line 190: should this just be `"extension"`? what does `stringify` do here?

rehandalal (approved):
nice work! lots of neat stuff in here. 

just one nit.

r+wc

bors d=@mythmon

webpack.config.js
mythmon: This needs to be the JS-source-code to replace `__ENV__` with. If it was just `"extension"`, then we'd end up with code like `if (__ENV__ === "extension")` being converted to `if (extension === "extension")`, which won't work.

This could be written as `__ENV__: '"extension"'`, but that looks more confusing to me.

rehandalal: ah! right makes sense.

rehandalal (approved):
bors r+

PR #455, opened 2020-09-29
Fix syntax error in generated files

### LINKED ISSUE #453: Syntax error in generated content.js on main
I noticed that the generated `content.js` has a syntax error, as reported by Firefox and Node. This shows up when I run `yarn watch` and also `yarn build`. I don't really understand it though. I don't think it's coming from any of our code, since ESLint isn't catching it, which implies it is part of our build process. Looking at the generated source code, I don't understand how the error *is* an error, and I don't understand how it got here.

I did a git bisect, and it points at either #451, or #442, depending on how I test. Neither of those make sense to me, since the `dotenv-webpack` bump looks unrelated to anything we do, and an update to `@testing-framework/react` shouldn't affect the build.

mythmon:
This was caused by a recent update to dotenv-webpack, in #451. I expect they are going to revert that change. For now though, we're going to revert the bump on our end, and add a CI check that would have caught it.

Fixes #453

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,9 @@ jobs:
       - run:
           name: Build extension
           command: yarn build
+      - run:
+          name: Syntax check built output
+          command: ls dist/*.js | xargs -l node --check
       - store_artifacts:
           path: web-ext-artifacts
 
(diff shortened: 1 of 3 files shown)


rehandalal (approved):
bors r+

PR #460, opened 2020-09-29
Show testing userIds for multi-pref-experiment branches

mythmon:
I've been sitting on this for a while, working on it in the background. These testing IDs are important to the recent QA process of testing specific branches of experiments, and having them easily accessible is very nice.

The testing ID shows up for each branch:

image

---

And also generally for the experiment (as it did before):

image


content/components/recipes/details/arguments/MultiPreferenceExperiment.js
```diff
@@ -85,6 +90,29 @@ export default function MultiPreferenceExperiment({ data }) {
                       <code>{branch.ratio}</code>
                     </div>
                   </div>
+                  <div className="flex-basis-0 flex-grow-1">
+                    <div className="d-flex align-items-center">
+                      <div className="pr-2 font-weight-bold">
+                        Testing clientId
+                        <HelpIcon>
+                          Setting the preference{" "}
```
rehandalal, line 98: nit: i found this a little confusing `{" "}`. It should probably be `&nbsp;`

rehandalal (changes requested):
r+wc

content/components/recipes/details/arguments/MultiPreferenceExperiment.js
mythmon: This is how prettier formats text when it wraps JSX lines. I tested and Prettier will leave `&nbsp;` if I put it there, but I think the automatic thing is easier.

rehandalal: oh right, this is all coming back to me.

rehandalal (approved):
bors r+

PR #461, opened 2020-09-30
Pending approval requests on the overview page

### LINKED ISSUE #403: In recipe overview, highlight recipes that need review
The recipe overview page should highlight recipes that have an open approval request. This recipes need attention from operators. This information should not be directly actionable however. Operators should have to open the recipe and review the changes.

tiftran:
fixes #403


mythmon:
I don't think we want to be doing anything that has to look at an unfiltered list of all the recipes on the server. I think we need to either have a way to limit it down to some subset, or more likely we'll need to add a filter on the server that can query by the state of approval.

content/components/overview/PendingReviews.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Link } from "react-router-dom";
+import { Col, List, Panel, Row, Tag } from "rsuite";
+
+import { RecipeV3 } from "devtools/types/recipes";
+import { chunkBy } from "devtools/utils/helpers";
+
+export const PendingReviews: React.FC<any> = (props: {
+  data: Array<RecipeV3>;
+}) => {
+  const { data } = props;
```
mythmon: You can say this a bit more concisely (and type safely) by putting the types in the generic slot:

```suggestion
export const PendingReviews: React.FC<{ data: Array<RecipeV3> }> = ({ data }) => {
```

content/components/overview/PendingReviews.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const pendingReviews = data.filter((recipe) => {
+    const {
+      latest_revision: { approval_request },
+    } = recipe;
+    return approval_request && approval_request.approved === null;
```
mythmon: We can use fancy things:

```suggestion
    return recipe.latest_revision.approval_request?.approved === null;
```

content/components/pages/OverviewPage.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import {
+  useEnvironmentState,
+  useSelectedNormandyEnvironmentAPI,
+} from "devtools/contexts/environment";
+import { PendingReviews } from "../overview/PendingReviews";
+import { Loader } from "rsuite";
+
+export const OverviewPage: React.FC<any> = ({}) => {
```
mythmon: In general, we shouldn't be using `any` in Typescript. It turns off most of the type safety. In this case since you don't take any props, you can just use the default value for the `FC` type parameter.

```suggestion
export const OverviewPage: React.FC = () => {
```

content/components/pages/OverviewPage.tsx
```diff
@@ -0,0 +1,52 @@
...
+    let recipeList = [];
+
+    if (approvalRequests.length) {
+      await Promise.all(
+        approvalRequests.map(async (approvalRequest) => {
+          const {
+            revision: { recipe_id },
+          } = approvalRequest;
+          const recipe = await normandyApi.fetchRecipe(recipe_id);
+          recipeList = [...recipeList, recipe];
```
mythmon: Since everything is local, `recipeList` wasn't passed in as an argument, and we don't need to make this change its identify any more than `let recipeList` already does, this can be the simpler version:

```suggestion
          recipeList.push(recipe);
```

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    jest
+      .spyOn(NormandyAPI.prototype, "fetchApprovalRequests")
+      .mockImplementation(() =>
+        Promise.resolve([pendingApprovalRequest]),
+      );
+    jest
+      .spyOn(NormandyAPI.prototype, "fetchRecipe")
+      .mockImplementation(() => Promise.resolve(recipe));
+
+    /* global renderWithContext */
```
mythmon: I think that if you rebase on the current main branch, this won't be needed anymore (though let me know if I'm wrong about that, if so I have more work to do...)

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    // @ts-ignore
+    const { getByText } = renderWithContext(<OverviewPage />);
+    await waitForElementToBeRemoved(getByText(/Loading Overview/));
```
mythmon, line 39: TIL you can use regex here. That's super useful!

content/types/recipes.ts
```diff
@@ -50,6 +50,13 @@ export interface ApprovalRequest {
   comment: string | null;
   created: string;
   creator: User;
+  revision: RevisionApprovalRequest;
+
```
mythmon: Nit: a blank line here.

mythmon (changes requested):

content/components/overview/PendingReviews.tsx
tiftran: oooh fancy

mythmon (approved):
Looks good, thanks!

bors r+

PR #466, opened 2020-10-01
Fix #404: fix press enter in address bar

### LINKED ISSUE #404: In-page address bar is not editable

leplatrem:
Fix #404

- code would crash if match pattern is `null`
- pattern wouldn't match `web+normandy://`

### DIFF
--- a/content/components/common/AppHeader.js
+++ b/content/components/common/AppHeader.js
@@ -72,8 +72,10 @@ function AddressBar() {
 
   const handleKeyPress = (ev) => {
     if (ev.key === "Enter") {
-      const addressMatch = address.match(/ext\+normandy:\/(.+?)$/);
-      history.push(addressMatch[1]);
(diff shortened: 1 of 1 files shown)


content/components/common/AppHeader.js
```diff
@@ -72,8 +72,10 @@ function AddressBar() {
 
   const handleKeyPress = (ev) => {
     if (ev.key === "Enter") {
-      const addressMatch = address.match(/ext\+normandy:\/(.+?)$/);
```
mythmon, line 75: Oops, this is my fault. I searched for `ext+normandy`, but forgot that some cases might have it escaped like this.

mythmon (approved):
bors r+

Issue #414, opened 2020-09-11
Skip suitability checks in web-mode


mythmon:
Suitability checking relies on using extension super powers. We can't use those powers when NDT is running as a webpage. We should not show suitability badges when in web-mode.

PR #470, opened 2020-10-02
hide run buttons for web

### LINKED ISSUE #415: Hide run buttons in web-mode
Running recipes needs extension super powers which we won't have in web-mode. We should hide all run buttons, including custom run, in web-mode.

tiftran:
fixes #415


mythmon (commented):
This looks like it would work well.

One thing I've been pondering is if we should have some higher order component for this and other conditional display stuff. Maybe like 

```js
<OnlyIf extension>
  <Dropdown.Item onSelect={this.handleRunButtonClick}>
  ...
</OnlyIf>
```

I'm not sure if that would work as well with Webpack's dead code removal though, so maybe not. Though I'm not sure if dead code removal even works in JSX in the first place. What do you think about that?

Also, I noticed this is marked as a draft. What else were you planning to change here?

tiftran:
I had left it on draft since I was missing tests. Having a high order component, this would mean hiding the 
the if `__ENV__ === "extension"` logic into a component right?

mythmon:
> Having a high order component, this would mean hiding the the if `__ENV__ === "extension"` logic into a component right?

Right. Also to be clear I'm asking for feedback about this idea, with the possibility of someone eventually implementing it. I'm not saying we should block this PR on adding that feature (unless you want to do it that way).

tiftran:
>  Also to be clear I'm asking for feedback about this idea, with the possibility of someone eventually implementing it

I was planning/hoping to probably pick up #414 next, maybe I'll let this idea sit with me while doing that?

Given the above, I think this PR is ready to be reviewed

mythmon (approved):
Sounds like a good plan.

bors r+

PR #478, opened 2020-10-07
Make VPN check work for web mode

### LINKED ISSUE #416: Change VPN check to work in web-mode
The VPN check currently relies on manipulating Firefox's connection cache via extension super powers. We won't be able to do this from the web mode version of NDT.

At the least, we should skip the connection-closing part of the VPN check when running in web-mode. It would be ideal to find a replacement to avoid having to hard refresh the page, but it shouldn't be a blocker.

rehandalal:
Fixes #416.  

I couldn't figure out a way to make this work without the refresh on web. I also considered adding an interval based check but I wasn't sure that would be useful.


mythmon (approved):
bors d+

rehandalal (commented):
bors r+

PR #479, opened 2020-10-07
hide suitability tag for web

### LINKED ISSUE #414: Skip suitability checks in web-mode

tiftran:
fixes #414


content/components/recipes/RecipeListing.tsx
```diff
@@ -149,17 +152,15 @@ class RecipeListing extends React.PureComponent<
...
-              <Dropdown.Item onSelect={this.handleRunButtonClick}>
-                Run
-                {running ? <Icon spin icon="reload" /> : <Icon icon="play" />}
-              </Dropdown.Item>
-              <Dropdown.Item onSelect={this.handleCustomRunClick}>
-                Custom Run <Icon icon="gear" />
-              </Dropdown.Item>
-            </>
-          )}
+          <OnlyIf type="extension">
```
mythmon: Ok, I apologize for this review, but I don't think that `OnlyIf` is actually going to work the way I hoped it would :speak_no_evil:. With the previous inline version, webpack was smart enough to evaluate the condition at build time, and then completely elide the  enclosed code and the `Suitabilities.tsx` file. With `OnlyIf`, it isn't clever enough to do that, and so we end up including all that unused code.

It isn't _really_ a problem since bundle size isn't something we care about a lot, but I think that the extra readability isn't worth the extra weight :frowning_face:.

I probably should have tested this before having you go off and do it, sorry.

tiftran: lol, it's all good

mythmon (approved):
Looks good, you just need to remove the lingering `OnlyIf` import that linting is complaining about.

bors d+

PR #480, opened 2020-10-08
Generate version JSON data for web build

### LINKED ISSUE #477: Add a version.json file for Dockerflow compatibility
Cloud Services expects hosted projects to include a `version.json` file that is placed at `__version__` in the hosted output. We should generate this at build time, so that we can upload it to s3 when we start deploying the web version of NDT.

The spec for the JSON object is here: https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md

rehandalal:
Fixes #477 

r?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,6 +84,16 @@ jobs:
           name: Build
           command: yarn build
       - run:
+          name: Create version JSON file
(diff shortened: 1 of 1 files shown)


mythmon (approved):
bors r+

PR #486, opened 2020-10-08
Hide logo from tests

### LINKED ISSUE #424: Logo SVG makes many tests hard to debug
Often while debugging test output, the testing framework tries to print the document in which it could not find the expected element. Most tests render the entire document (so it doesn't have to mock state as much), which means that the document includes the SVG logo, which is lengthy.

image

The SVG content extends a long ways and eventually the test framework just cuts it off, usually hiding the place where the bug actually is. This makes testing more difficult than it should be.

rehandalal:
Fixes #424 

r?

### DIFF
--- a/content/tests/components/common/HideFromTests.test.tsx
+++ b/content/tests/components/common/HideFromTests.test.tsx
@@ -1,0 +1,15 @@
+import { render } from "@testing-library/react";
+import React from "react";
+
+import { HideFromTests } from "devtools/components/common/HideFromTests";
+
+describe("HideFromTests", () => {
+  it("should hide it's children", async () => {
+    const { queryByText } = render(
+      <HideFromTests>
+        <div>SHOW ME!</div>
+      </HideFromTests>,
+    );
+    expect(queryByText("SHOW ME!")).toBeNull();
+  });
+});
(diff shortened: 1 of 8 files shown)


leplatrem (approved):
LGTM

rehandalal:
bors r=@leplatrem

content/tests/components/common/HideFromTests.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import React from "react";
+
+import { HideFromTests } from "devtools/components/common/HideFromTests";
+
+describe("HideFromTests", () => {
+  afterEach(async () => {
+    await cleanup();
+  });
+
+  it("should hide it's children", async () => {
```
mythmon, line 11: I think we should have the opposite test as well, since we can manipualte the globals at test time. You can see how Tif messed with `__ENV__` for tests here: https://github.com/mozilla-extensions/normandy-devtools/pull/479/files#diff-2049bef6a60a06f2ebc6de2ec30167d5R78

mythmon: Oh, I was too slow. Oh well

rehandalal: I'll add another test in a new PR

PR #489, opened 2020-10-08
Make sure typescript files get checked by Therapist

### LINKED ISSUE #487: Enable ESLint checks for typescript files in therapist
Looks like we're not running ESLint on typescript files in Therapist and as a result in CI etc.

We should add `.ts` and `.tsx` to the files passed through to eslint in `.therapist.yml`.

mythmon:
Fixes #487


rehandalal (approved):
thanks!

bors r+

PR #499, opened 2020-10-13
Ending overview

### LINKED ISSUE #402: In recipe overview, highlight recipes that are due to end
Integrating data from Experimenter, we should highlight recipes that scheduled to end. If the proposed end time is in the past or in the near future there should be a visual indication that the recipe needs attention, possibly by pulling it into a dedicated section.

Additionally, this change should be actionable. An operator should be able to easily disable the recipe without leaving the page.

tiftran:
fixes #402

### DIFF
--- a/content/components/overview/EndingRecipes.tsx
+++ b/content/components/overview/EndingRecipes.tsx
@@ -1,0 +1,112 @@
+import React, { ReactElement } from "react";
+import { Link } from "react-router-dom";
+import { Alert, Col, Icon, IconButton, Panel, Row, Tag } from "rsuite";
+
+import { useSelectedNormandyEnvironmentAPI } from "devtools/contexts/environment";
+import { RecipeV3 } from "devtools/types/recipes";
+import { chunkBy } from "devtools/utils/helpers";
+
+export const EndingRecipes: React.FC<{
+  data: Array<{ endDate: number; recipe: RecipeV3 }>;
+}> = ({ data }) => {
+  const sevenDaysFuture = new Date();
+  sevenDaysFuture.setDate(sevenDaysFuture.getDate() + 7);
+
+  const renderEndingSoonListItem = (): Array<ReactElement> => {
+    const endingRecipes = data.filter((recipeData) => {
+      const endingDate = new Date(recipeData.endDate);
+      return endingDate <= sevenDaysFuture;
+    });
+    endingRecipes.sort((exp1, exp2) => {
+      return exp1.endDate - exp2.endDate;
+    });
+    return chunkBy(endingRecipes, 3).map((recipeChunk, rowIdx) => {
+      return (
+        <Row key={rowIdx}>
+          {recipeChunk.map((recipeData, colIdx) => {
+            return (
+              <Col key={`col-${colIdx}`} md={8} sm={24}>
+                <EndingRecipeCard recipeData={recipeData} />
+              </Col>
+            );
+          })}
+        </Row>
+      );
+    });
+  };
+
+  return (
+    <>
+      <h3>Ending Recipes</h3>
+      {renderEndingSoonListItem()}
+    </>
+  );
+};
+
+const EndingRecipeCard: React.FC<{
+  recipeData: { endDate: number; recipe: RecipeV3 };
+}> = ({ recipeData }) => {
+  const { endDate, recipe } = recipeData;
+  const [isButtonLoading, setIsButtonLoading] = React.useState(false);
+  const normandyApi = useSelectedNormandyEnvironmentAPI();
+
+  const handleDisableClick = async (): Promise<void> => {
+    setIsButtonLoading(true);
+    try {
+      await normandyApi.disableRecipe(recipe.id);
+      Alert.success("Recipe Successfully disabled");
+    } catch (err) {
+      console.warn(err.message, err.data);
+      Alert.error(`An Error Occurred: ${JSON.stringify(err.message)}`, 5000);
+    } finally {
+      setIsButtonLoading(false);
+    }
+  };
+
+  const endingDate = new Date(endDate);
+  return (
+    <Panel
+      bordered
+      className="recipe-listing mb-2"
+      header={<CardHeader recipe={recipe} />}
+    >
+      <div className="d-flex flex-wrap m-0">
+        <div className="flex-grow-1">
+          <p>Experimenter End Date: </p>
+          {endingDate.toLocaleDateString()}
+        </div>
+
+        <IconButton
+          className="ml-1"
+          color="red"
+          disabled={!recipe.latest_revision.enabled}
+          icon={<Icon icon="close-circle" />}
+          loading={isButtonLoading}
+          onClick={handleDisableClick}
+        >
+          Disable
+        </IconButton>
+      </div>
+    </Panel>
+  );
+};
+
+const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({
+  recipe,
+}): ReactElement => {
+  return (
+    <>
+      <Link to={`recipes/${recipe.id}`}>
+        <Tag className="mr-half" color="violet">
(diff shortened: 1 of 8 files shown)


content/components/overview/EndingRecipeCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { useSelectedNormandyEnvironmentAPI } from "devtools/contexts/environment";
+import { RecipeV3 } from "devtools/types/recipes";
+
+export const EndingRecipeCard: React.FC<{
+  recipeData: { endDate: number; recipe: RecipeV3 };
+}> = ({ recipeData }) => {
+  const [isButtonLoading, setIsButtonLoading] = React.useState(false);
+  const normandyApi = useSelectedNormandyEnvironmentAPI();
+
+  const renderHeader = (recipe): ReactElement => {
```
mythmon: This is awfully close to just being its own component. Maybe move it to the top level and make it this?

```tsx
const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({ recipe }) => { ... }
```

It doesn't have to be in a separate file or anything, but I think treating it as an independent component is better.

content/components/overview/EndingRecipeCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+      Alert.success("Recipe Successfully disabled");
+    } catch (err) {
+      console.warn(err.message, err.data);
+      Alert.error(`An Error Occurred: ${JSON.stringify(err.message)}`, 5000);
+    } finally {
+      setIsButtonLoading(false);
+    }
+  };
+
+  const renderDisableButton = (recipe): ReactElement => {
```
mythmon: Do you find having this outside of the main JSX definition helpful? Personally I find this "Declare in bits and then assemble" method kind of disorienting, since I prefer to see the whole before the parts.

content/components/pages/OverviewPage.tsx
```diff
@@ -35,13 +41,35 @@ export const OverviewPage: React.FC = () => {
...
+            const recipe = await normandyApi.fetchRecipe(
+              experiment.normandy_id,
+            );
+            liveRecipes.push({ endDate: experiment.end_date, recipe });
+          }
+        }),
+      );
+    }
+
+    setLiveRecipes(liveRecipes);
```
mythmon: Does anything reset `liveRecipes`? Or do we just keep pushing to it, adding more things everytime this function gets called?

I think it would be better to have a local array to accumulate these. The below should work for that (though I typed in this comment, so maybe paste it into your editor to check for lint issues first):

```
    const experiments = await experimenterApi.fetchExperiments({
      status: "Live",
    });

	const newLiveRecipes = await Promise.all(
	  experiments.flatMap(async (experiment) => {
        if (experiment.normandy_id) {
          const recipe = await normandyApi.fetchRecipe(experiment.normandy_id);
          return [{ endDate: experiment.end_date, recipe }];
        }
        return [];
	}));

    setLiveRecipes(newLiveRecipes);
```

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -35,10 +41,67 @@ describe("OverviewPage", () => {
...
+
+    const recipe2 = recipeFactory.build({ latest_revision: { enabled: true } });
+
+    const recipe3 = recipeFactory.build({
+      latest_revision: { enabled: false },
+    });
+
+    const exp1 = experimenterResponseFactory.build({ normandy_id: recipe1.id });
+    const exp2 = experimenterResponseFactory.build({ normandy_id: recipe2.id });
+    const exp3 = experimenterResponseFactory.build({ normandy_id: recipe3.id });
```
mythmon: This works fine, you may be interested in the `buildMany` variant of the deafult factory though.

```tsx
const recipes = recipeFactory.buildMany([
	{ latest_revision: { enabled: true } },
	{ latest_revision: { enabled: true } },
	{ latest_revision: { enabled: false } },
]);
const experiments = experimentResponseFactory.buildMany(recipes.map((r) => ({ normandy_id: r.id })));
```

content/utils/experimenterApi.js
```diff
@@ -5,6 +5,12 @@ export default class ExperimenterAPI extends API {
     return new URL(`api/v${version}/`, this.environment.experimenterUrl).href;
   }
 
+  async request(options) {
```
mythmon: This seems like a no-op. Couldn't you remove it?

content/components/overview/EndingRecipeCard.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { ReactElement } from "react";
+import { Link } from "react-router-dom";
+import { Alert, Icon, IconButton, Panel, Tag } from "rsuite";
+
+import { useSelectedNormandyEnvironmentAPI } from "devtools/contexts/environment";
+import { RecipeV3 } from "devtools/types/recipes";
+
+export const EndingRecipeCard: React.FC<{
```
mythmon: Personally I'd put this and `EndingRecipes` in the same file; they aren't very useful independent of each other. I don't think we have a hard rule about one-component-per-file, though if we do then lets stick with it.

content/utils/experimenterApi.js
tiftran: tslint wasn't happy with me  when i had searchParams for https://github.com/mozilla-extensions/normandy-devtools/pull/499/files#diff-6d30f20f347f56d86c0c9ef1029997e54bb2713522f3dced5efb6ba0b90f13c7R29

wasn't sure if tehre was another way to make it happy

mythmon: The super method has JSDoc-style type annotations that don't mention `data`, so it makes sense that tslint was grumpy. Adding this interstitial method erases the type information, so it just turns off the checks. Better would be to add `data` as a possible parameter here:

https://github.com/mozilla-extensions/normandy-devtools/blob/ee192ff3d9b5786e766905921ee7b9d348ffecd2/content/utils/api.js#L19-L27

Without converting it over to Typescript (which would be better), you could probably just add

```js
 * @param {Object} [args.data] Data to convert JSON serialize into the querystring (for GET and HEAD requests) or the request body (for other methods).
```

content/components/pages/OverviewPage.tsx
tiftran: tried to use flatMap but with Promise.all it returned `[[], [{<insert recipe here>}], []...]`

mythmon: Oh, right. You'd have to flatten outside the promise. Something like 

```
(await Promise.all(
	experiments.map(async (exp) => { ... }))
).flat()
`.

 Does that work?

File: content/tests/factories/index.ts [hunk 7]
```diff
@@ -140,7 +140,11 @@ export default class Factory<T, O extends object = object>
   ): Array<T> {
     const rv: Array<T> = [];
     for (let i = 0; i < overrides.length; i++) {
-      rv.push(this._buildFunc(overrides[i], options[i]));
+      if (options) {
+        rv.push(this._buildFunc(overrides[i], options[i]));
+      } else {
+        rv.push(this._buildFunc(overrides[i]));
```
mythmon, line 146: Oh, good catch.

content/components/overview/EndingRecipes.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </IconButton>
+      </div>
+    </Panel>
+  );
+};
+
+const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({
+  recipe,
+}): ReactElement => {
+  console.log(recipe);
```
mythmon: This should probably be removed.

mythmon (approved):
This looks good, thanks for the revisions. There are one or two `console.log`'s floating around, and feel free to squash this if you want. Besides that it's good to go.

bors d+

PR #506, opened 2020-10-13
Show Experimenter branches in action details

leplatrem:


content/components/recipes/details/arguments/BranchedAddonStudy.js
```diff
@@ -4,31 +4,42 @@ import { Tag } from "rsuite";
...
+          ["slug", "description", "ratio", "extensionApiId"],
+          {
+            slug(index, value) {
+              return <code>{value}</code>;
+            },
+            description(index) {
+              const { slug } = data.arguments.branches[index];
+              return experimenterData.variants
+                ? experimenterData.variants[slug]
+                : null;
```
mythmon: This can done more concisely now with nullish coalescing and optional chaining:

```suggestion
              return experimenterData.variants?.[slug] ?? null;
```

content/less/main.less
```diff
@@ -203,3 +203,7 @@ code.hljs {
 .rs-form-vertical .rs-form-group .rs-input-group {
   width: auto;
 }
+
+.margin-right {
+  margin-right: 2em;
+}
```
mythmon: In `spacing.less`, we define a bunch of margin helpers. I believe this would be something like `mr-2` ("margin right, double size).

leplatrem:
I left the list of branches in the details panel of Experimenter, mainly because I figured that sometimes branches/slugs do not match (branches defined in Experimenter that are not present in Normandy arguments)

See [web+normandy://prod/recipes/1059](web+normandy://prod/recipes/1059)

mythmon (approved):
bors d+

PR #512, opened 2020-10-14
Fix webpack dev workflow

mythmon:


mythmon:
I didn't see something in the webpack-cli changelog that dealt with `webpack --config-name X`, which is where our problem is. Lets just revert this for now and maybe figure it out later.

rehandalal (approved):
bors r+

PR #519, opened 2020-10-15
added pausing recipes to overview

### LINKED ISSUE #401: In recipe overview, highlight recipes that are due to be paused
Integrating data from Experimenter, we should highlight recipes that scheduled to pause enrollment soon. If the proposed enrollment time is in the past or in the near future there should be a visual indication that the recipe needs attention, possibly by pulling it into a dedicated section.

Additionally, this change should be actionable. An operator should be able to easily pause the recipe without leaving the page.

tiftran:
fixes #401

### DIFF
--- a/content/components/overview/PausingRecipes.tsx
+++ b/content/components/overview/PausingRecipes.tsx
@@ -98,9 +98,7 @@ const PausingRecipeCard: React.FC<{
   );
 };
 
-const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({
-  recipe,
-}): ReactElement => {
+const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({ recipe }) => {
   return (
     <>
       <Link to={`recipes/${recipe.id}`}>
--- a/content/components/overview/PausingRecipes.tsx
+++ b/content/components/overview/PausingRecipes.tsx
@@ -107,8 +107,7 @@ const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({ recipe }) => {
         </Tag>
       </Link>
       <Link
-        // to get rid of blue text and underline
-        style={{ textDecoration: "inherit", color: "inherit" }}
+        className="text-reset text-decoration-none"
         to={`recipes/${recipe.id}`}
       >
         {recipe.latest_revision.name}
--- a/content/components/overview/PausingRecipes.tsx
+++ b/content/components/overview/PausingRecipes.tsx
@@ -107,6 +107,7 @@ const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({ recipe }) => {
         </Tag>
       </Link>
       <Link
+        // to get rid of blue text and underline
         style={{ textDecoration: "inherit", color: "inherit" }}
         to={`recipes/${recipe.id}`}
       >
--- a/content/components/overview/PausingRecipes.tsx
+++ b/content/components/overview/PausingRecipes.tsx
@@ -1,0 +1,119 @@
+import React, { ReactElement } from "react";
+import { Link } from "react-router-dom";
+import { Alert, Col, Icon, IconButton, Panel, Row, Tag } from "rsuite";
+
+import { useSelectedNormandyEnvironmentAPI } from "devtools/contexts/environment";
+import { RecipeV3 } from "devtools/types/recipes";
+import { chunkBy } from "devtools/utils/helpers";
+
+export const PausingRecipes: React.FC<{
+  data: Array<{ pauseDate: number; recipe: RecipeV3 }>;
+}> = ({ data }) => {
+  const sevenDaysFuture = new Date();
+  sevenDaysFuture.setDate(sevenDaysFuture.getDate() + 7);
+
+  const renderPausingSoonListItem = (): Array<ReactElement> => {
+    const pausingRecipes = data.filter((recipeData) => {
+      const pausingDate = new Date(recipeData.pauseDate);
+      return pausingDate <= sevenDaysFuture;
+    });
+    pausingRecipes.sort((exp1, exp2) => {
+      return exp1.pauseDate - exp2.pauseDate;
+    });
+    return chunkBy(pausingRecipes, 3).map((recipeChunk, rowIdx) => {
+      return (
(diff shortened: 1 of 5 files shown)


content/components/overview/PausingRecipes.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { RecipeV3 } from "devtools/types/recipes";
+import { chunkBy } from "devtools/utils/helpers";
+
+export const PausingRecipes: React.FC<{
+  data: Array<{ pauseDate: number; recipe: RecipeV3 }>;
+}> = ({ data }) => {
+  const sevenDaysFuture = new Date();
+  sevenDaysFuture.setDate(sevenDaysFuture.getDate() + 7);
+
+  const renderPausingSoonListItem = (): Array<ReactElement> => {
```
mythmon, line 15: I expect that taking this out of the inner function and integrating it into its parent would result in easier to follow logic, but it's fine either way.

content/components/overview/PausingRecipes.tsx
```diff
@@ -0,0 +1,119 @@
...
+          Pause
+        </IconButton>
+      </div>
+    </Panel>
+  );
+};
+
+const CardHeader: React.FC<{ recipe: RecipeV3 }> = ({
+  recipe,
+}): ReactElement => {
```
mythmon: You shouldn't need to separately declare the return value, since `React.FC` already says that this function will return a react element.

content/components/overview/PausingRecipes.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <>
+      <Link to={`recipes/${recipe.id}`}>
+        <Tag className="mr-half" color="violet">
+          {recipe.id}
+        </Tag>
+      </Link>
+      <Link
+        style={{ textDecoration: "inherit", color: "inherit" }}
```
mythmon: It's not obvious what this is doing here. Can you add a comment explaining what's going on?

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -100,4 +100,65 @@ describe("OverviewPage", () => {
 
     expect(NormandyAPI.prototype.disableRecipe).toBeCalled();
   });
+
+  it("should display need to pause recipes", async () => {
+    const recipes = recipeFactory.buildMany([{}, {}, {}]);
```
mythmon: If you don't care to specify any overrides, you can use the even simpler `const recipes = recipeFactory.buildCount(3);`

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -100,4 +100,65 @@ describe("OverviewPage", () => {
...
+      .spyOn(NormandyAPI.prototype, "fetchApprovalRequests")
+      .mockImplementation(() => Promise.resolve([]));
+    jest
+      .spyOn(NormandyAPI.prototype, "fetchRecipe")
+      .mockReturnValueOnce(Promise.resolve(recipes[0]))
+      .mockReturnValueOnce(Promise.resolve(recipes[1]))
+      .mockReturnValueOnce(Promise.resolve(recipes[2]))
+      .mockReturnValueOnce(Promise.resolve(recipes[0]))
+      .mockReturnValueOnce(Promise.resolve(recipes[1]))
+      .mockReturnValueOnce(Promise.resolve(recipes[2]));
```
mythmon: I expect you could do this with a loop.

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -100,4 +100,65 @@ describe("OverviewPage", () => {
...
+
+    fireEvent.click(pauseButton[0]);
+
+    expect(NormandyAPI.prototype.patchRecipe).toBeCalledWith(recipes[1].id, {
+      comment: "One-click pause",
+      arguments: {
+        ...recipes[1].latest_revision.arguments,
+        isEnrollmentPaused: true,
+      },
+    });
```
mythmon: This should probably be a separate test, since it is a mostly independent set of functionality.

mythmon (changes requested):

content/components/overview/PausingRecipes.tsx
tiftran: ohhh, yes i was gonna highlight this when I initially had this PR up for review...
I was trying to get rid of the ugly blue and underlining, probs not the way to go about it

content/tests/components/pages/OverviewPage.test.tsx
```diff
@@ -100,4 +102,93 @@ describe("OverviewPage", () => {
...
+
+    const mockFetchRecipe = jest
+      .fn()
+      .mockImplementation((i): RecipeV3 => recipes[i % 3]);
+
+    jest
+      .spyOn(NormandyAPI.prototype, "fetchRecipe")
+      .mockImplementation(() =>
+        Promise.resolve(mockFetchRecipe(mockFetchRecipe.mock.calls.length)),
+      );
```
tiftran, line 129: is this better than having 6 `mockReturnValueOnce`s or is this still considered clunky?

mythmon: Huh. I hadn't thought about using the call length itself. That seems clever. +1 to this.

content/components/overview/PausingRecipes.tsx
mythmon: Oh I meant a comment in the code, so we aren't confused in the future. :) I don't have any ideas for how to get rid of it either. May @rdalal does, but I think with a comment added here it will be fine.

mythmon (changes requested):

rdalal:
Just letting you know---I think you have the wrong rdalal.

mythmon:
Oh, sorry for the bother. I meant @rehandalal.

content/components/overview/PausingRecipes.tsx
rehandalal: my preference for this kind of thing would be to create a class that can be re-used places we want to do this. maybe something like `.link-inherit-color-decoration`.

mythmon (approved):
bors r+

PR #526, opened 2020-10-17
added search functionality to rollback dropdown

tiftran:


content/components/recipes/form/arguments/PreferenceRollback.tsx
```diff
@@ -19,12 +19,30 @@ export default function PreferenceRollback() {
...
+  }> = rollouts.map((rollout) => ({
+    label: `${rollout.id}: ${rollout.latest_revision.name} `,
     value: rollout.latest_revision.arguments.slug,
   }));
 
+  const rolloutMenu = (label, item): ReactElement => {
+    return (
+      <div>
+        {label}
+        <small style={{ opacity: 0.5, fontSize: "95%" }}>{item.value}</small>
```
mythmon: I think this should be `<small className="text-subtle">` instead?

mythmon (changes requested):

content/components/recipes/form/arguments/PreferenceRollback.tsx
tiftran: you're right, i'm such a goof 🙈

mythmon (approved):
bors r+

PR #532, opened 2020-10-19
Revert "Bump webpack-cli from 3.3.12 to 4.0.0"

rehandalal:
#496 introduced an error that I wasn't able to easily fix so I figured we should just revert this for now.

r?


mythmon:
#512 should cover this already, and has a test that would have caught it.

PR #497, opened 2020-10-12
Only render the inner page for RecipeDetails tests

mythmon:
I was having a really hard time with tests on #462, and I found that making these tests more focused on just the page in question made them easier to debug and more resilient to unrelated changes. I extracted it into a separate PR because the other one was touching too many things.

PR #462, opened 2020-09-30
Search and filtering UI

mythmon:
- [x] Search, filter by enabled, and filter by action
- [x] Fetch action list from server instead of hard coding it
- [x] Link recipe query to the URL
- [x] Tests

PR #533, opened 2020-10-19
Show correct status for recipe on details page

rehandalal:
Fixes the issue with the details page showing the wrong status when a draft revision exists. 

Also fixes a problem where the enable button is shown for enabled recipes after a new approval request is approved.

r?

### DIFF
--- a/content/contexts/recipeDetails.js
+++ b/content/contexts/recipeDetails.js
@@ -9,6 +9,7 @@ export const INITIAL_RECIPE_DATA = {
 
 const initialState = {
   data: INITIAL_RECIPE_DATA,
+  statusData: INITIAL_RECIPE_DATA,
   importInstructions: "",
   clientErrors: {},
 };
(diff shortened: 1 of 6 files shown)


content/contexts/recipeDetails.js
```diff
@@ -26,7 +27,8 @@ function reducer(state, action) {
     case ACTION_UPDATE_DATA:
       return {
         ...state,
-        data: action.data,
+        data: action.data || state.data,
+        statusData: action.statusData || state.statusData,
```
mythmon, line 31: Is this guarding against the passed in `data` and `statusData` being undefined/null? When does that happen? Would this be better as `??` instead of `||`?

rehandalal: I suppose `??` would make sense here. It could be undefined/null if the action passed to dispatch contained only one of these.

mythmon: I looked around at where `ACTION_UPDATE_DATA` is used, and I see that there are a lot of times that we would only have one of the fields, especially because the editing forms seem to share this context. Alright.

mythmon (approved):
bors r+

mythmon:
This is good to go, but it got hit by some fragile tests. I think that landing #462 and #497, and then rebasing this will fix the test failures.

PR #543, opened 2020-10-21
Fix #405: Add iframe with enrollment dashboard

### LINKED ISSUE #405: Pull in Grafana-like data
We should have a data system that can get data about the enrollment and population of experiments, like used on [the Grafana live monitoring dashboard](https://grafana.telemetry.mozilla.org/d/XspgvdxZz/experiment-enrollment?orgId=1&var-experiment_id=bug-1629108-pref-password-import-autocomplete-experiment-release-77-78&from=1590969600000&to=1595548800000) that Experimenter uses.

We may have to do something tricky with authentication here, so it's worth getting just data fetching working, separately from any display.

leplatrem:
Screenshot_2020-10-21 Normandy Devtools

- [x] POC
- [ ] Decide whether we embed panels or single dashboard (I favor the latter for maintainability)
- [ ] Prompt for snapshot ID
- [ ] Integrate with environment switch
- [ ] Get Grafana feature deployed in prod

### DIFF
--- a/content/components/recipes/details/EnrollmentDetails.tsx
+++ b/content/components/recipes/details/EnrollmentDetails.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+
+import CollapsibleSection from "devtools/components/recipes/details/CollapsibleSection";
+import { useRecipeDetailsData } from "devtools/contexts/recipeDetails";
+
+const EnrollmentDetails: React.FunctionComponent = () => {
+  const server = "https://stage.grafana.nonprod.dataops.mozgcp.net";
+  const snapshot = "QZ3X0hpzQ67uBbvbPD02cE5Wqa0yRCar";
+  const { experimenter_slug: slug } = useRecipeDetailsData();
+
+  const url = `${server}/dashboard/snapshot/${snapshot}?var-experiment_id=${slug}`;
+
+  return (
+    <CollapsibleSection
+      collapsed={true}
+      headerButtons={<></>}
+      title={<h6>Enrollment Details</h6>}
+    >
+      <div className="py-1 pl-4">
+        <div className="mt-4">
+          {slug ? <iframe height="800" src={url} width="100%"></iframe> : "N/A"}
+        </div>
+      </div>
+    </CollapsibleSection>
+  );
+};
+
+export default EnrollmentDetails;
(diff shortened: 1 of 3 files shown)


content/components/recipes/details/EnrollmentDetails.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+import CollapsibleSection from "devtools/components/recipes/details/CollapsibleSection";
+import { useRecipeDetailsData } from "devtools/contexts/recipeDetails";
+
+const EnrollmentDetails: React.FunctionComponent = () => {
+  const server = "https://stage.grafana.nonprod.dataops.mozgcp.net";
+  const snapshot = "QZ3X0hpzQ67uBbvbPD02cE5Wqa0yRCar";
```
mythmon, line 8: This looks suspiciously key-like. Can you add a comment explaining what it is for, and if it is safe explain why?

mythmon (commented):
Are you also planning to adapt the visible area of the graph to the experiment at all? Or just stick with the defaults?

content/components/recipes/details/EnrollmentDetails.tsx
leplatrem: It's the snapshot id, but I don't think it's safe to make it public, unless we have a clear confirmation that enrollment data can be publicly accessible.
I have to confirm with Blake whether snapshots can be embedded anywhere or just under our domains. 
Otherwise I was thinking of prompting and storing it in localstorage for example

leplatrem:
> Are you also planning to adapt the visible area of the graph to the experiment at all? 

I didn't understand that. 

I was thinking of embedding the whole dashboard pointed at the current recipe (with field `experiment_id` filled). This way the dashboard is defined once for everything, and can evolve over time without re-releasing (for ex. add the enrollment/unenrollment rate)

mythmon:
Ah sorry, I was talking about the visible time period. For example, Experimenter adjusts the bounds of the graph to the duration of the experiment (with some extra details to make sure all the data is visible). I could see an argument for this, or for always viewing a recent period of time.

content/components/recipes/details/EnrollmentDetails.tsx
mythmon: Well, it's public now. If it isn't safe to share we may want to revoke it.

We don't yet have an answer about making enrollment data public.

leplatrem: Yeah, but don't worry, it's only for this only test dashboard on the stage grafana instance, nothing to see there. This PR was just a POC for embedding iframes without auth :)

leplatrem:
I may have failed here. 

Even if their value is shown in the UI, passing the variables via the URL has no effect on the plotted data :sob:

PR #551, opened 2020-10-22
Code split less aggressively, to fix Codemirror

mythmon:
Before, we split any chunk that could be split. With this PR we revert to Webpack's default of only splitting async chunks, which is safer. This in theory produces more overall code, since things can be duplicated between chunks. That's preferable to a broken site though.

image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -233,7 +233,6 @@ function makeWebConfig(development) {
 
     optimization: {
       minimize: !development,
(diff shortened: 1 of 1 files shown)


mythmon:
@rehandalal did you find a way to make Codemirror more compatible with code splitting? If not lets use this as a fallback.

rehandalal (approved):
i can't find anything immediate. let's start here to unbreak things and then explore further

bors r+

PR #552, opened 2020-10-22
Make all codecov checks "informational" instead of failing CI

mythmon:
The codecov coverage reports are very useful, but flagging PRs as failing CI is noise that slows things down, at least for me. I think that the comments in the PRs are enough to provide the test coverage nudge we wanted.


rehandalal (approved):
bors r+

PR #568, opened 2020-10-27
Ref #405 - Show link to Telemetry

leplatrem:


content/components/recipes/details/DetailsHeader.js
```diff
@@ -147,6 +149,24 @@ export default function DetailsHeader() {
...
+      status,
+    } = useExperimenterDetailsData();
+    telemetryLink = (
+      <TelemetryLink
+        endDate={endDate}
+        normandySlug={normandySlug}
+        startDate={startDate}
+        status={status}
+      />
+    );
```
mythmon: If it's ok to pass extra things into the component (and with Typescript that works better than without), you could do this much more concisely:

```suggestion
    telemetryLink = <TelemetryLink {...useExperimenterDetailsData()} />;
```

mythmon (approved):

mythmon:
Oh well that didn't work, but it didn't work because the previous one was actually type-unsafe in a way that Typescript couldn't see until now. 

> content/components/recipes/details/DetailsHeader.js(154,22): error TS2322: Type '{ publicDescription: string; proposedStartDate: Date; proposedEndDate: Date; proposedDuration: number; startDate?: Date; endDate?: Date; variants: Record<string, string>; status: string; normandySlug: string; }' is not assignable to type 'TelemetryLinkProps'.
>
> Property 'endDate' is optional in type '{ publicDescription: string; proposedStartDate: Date; proposedEndDate: Date; proposedDuration: number; startDate?: Date; endDate?: Date; variants: Record<string, string>; status: string; normandySlug: string; }' but required in type 'TelemetryLinkProps'.

Or more concisely: `startDate` and `endDate` are optional in Experimenter data, but are said to be required in `TelemetryLink`.

leplatrem:
> Or more concisely: startDate and endDate are optional in Experimenter data, but are said to be required in TelemetryLink.

Hmm, yeah, there is an implicit dependency between `experimenter_slug` and the rest of Experimenter fields. 

If you think there's a cleaner way to express that using available types and directives, I'd be happy to adjust. 
We could add an issue to improve tests/behaviour of the `RecipeDetailsPage` context when recipe doesn't have or hasn't loaded the experiment details yet.

mythmon:
I'm not sure I follow. The type returned by `useExperimenterDetailsData` is an `ExperimenterData`, which has the properties `{startDate?: Date, endDate?: Date}`. The props for `TelemetryLink` had `{startDate: Date, endDate: Date}`. However, `TelemetryLink` did properly expect that those might be missing. Simply marking them as optional, as I did in my commit, fixes the issue.

Now, if you wanted to say that field A depends on field B in general, there is a way to do that. In my experience it often ends up complicating other things:

```typescript
type Book = 
	| { title: string; author?: undefined }
	| { title: string; author: string; authorBirthYear: number }
```

You can see it in action [on the Typescript Playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQg9nA1lAvFAUASAD5QN5TACWwANhAFxQDOwATkQHYDmA3FAIYCuwAFnHQD8VLowAmEAGZMIYqAF8suAsTKUa9Jm049+dKrQYt23PgJhE6fAJoQO+qIy4BbAEYQ6C1unQB6X1BI6OTAUK4AjFTwSKj4hCTkVABEAKqMquRiSV5+AR50AsEQoa4ATFEIyGgqCepJABJEzLxQAOpEAF72YnQgSQA0OmYOSQAiRByM0KNckxDZ8t7+gYhFJQDMFTHV8WrJjc1tnd29A0N6yeNzUDNzZ6Z6Fla8tvZU4QCcAKylCj7LABVeERqFAxHAINRGAByUIAdwEiEGrh4UAB4EhAGMGGBQuDITDQohGHA4YReBxgGswgAWLZVKAAKQAygB5AByADowPZqBAABTQznC6EASm86FcNM5GWgywgAA8QcBqJLpQ8BFB5UraNRkajnBwQGFoKIJNIpmI1ZyNXQnjY7J55XQCnRBvjQSTQorlT4iJIoPypTbdAJRfh0FAo7SQ8MtXkdSrBiDOBojMwMNGY7b7S9HextcrVfIgA).

leplatrem:
> Simply marking them as optional, as I did in my commit, fixes the issue.

Yes, makes sense!

> Now, if you wanted to say that field A depends on field B in general, there is a way to do that

It was late yesterday and couldn't make myself clear. I was more referring to the coupling there is between a truthy `recipeData.experimenter_slug` and the experimenter context not being empty. Specifically this part:

https://github.com/mozilla-extensions/normandy-devtools/blob/2789495b3d52cb43279e3ef252df4693371be56d/content/components/pages/RecipeDetailsPage.tsx#L31-L60

mythmon:
Ahh, I see. That makes more sense now. I don't think there is much we need to do at the type level to express that, and I'm happy with the current state of testing. We are testing the positive case, and I think the negative case is tested by all the other tests not breaking. I'm not concerned about it.

Issue #534, opened 2020-10-20
Test output is extremely noisy


rehandalal:
The test output is littered with info, warnings and errors.

I think we should hide all the `console.info` output and address all the warnings and errors. Also we should probably make it so that the tests fail if there are any warnings or errors so we are forced to fix them as they come up.

PR #583, opened 2020-11-02
Enable eslint-plugin-react-hooks

mythmon:
This is a recommended-by-React part of using hooks that we haven't been doing. As such, there are a bunch of errors that showed up. Some of these were problems in practice, and showed up in the console during testing.

Some of the problems pointed out by the linter were false positives, because of our build-time invariants. Others didn't seem wrong to me, but fixing them the way the linter suggested broke our tests. I left comments on those, and maybe we can figure out why in the future.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,6 +57,7 @@ module.exports = {
     "plugin:prettier/recommended",
     "plugin:mozilla/recommended",
     "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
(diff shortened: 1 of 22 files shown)


mythmon:
Let's wait to merge this until #584 is done.

mythmon:
I'm going to close this for now. With the testing changes that we are working on for #534, it will be hard to merge this in after-the fact. It's definitely something we should do in the future though.

PR #585, opened 2020-11-03
Set up NDT web deployments

mythmon:
This sets up CircleCI to deploy the website on tags, and makes a new tagged release that we can use to trigger the deploy once this merges.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,20 +84,12 @@ jobs:
           name: Build
           command: yarn build
       - run:
-          name: Create version JSON file
-          command: |
-            printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
-              "$CIRCLE_SHA1" \
-              "${CIRCLE_TAG-}" \
-              "$CIRCLE_PROJECT_USERNAME" \
-              "$CIRCLE_PROJECT_REPONAME" \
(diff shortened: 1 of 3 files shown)


rehandalal (approved):
bors r+

.circleci/config.yml
```diff
@@ -140,6 +131,47 @@ jobs:
...
+        type: env_var_name
+      bucket-name:
+        type: env_var_name
+      distribution-id:
+        type: env_var_name
+      secret-access-key:
+        type: env_var_name
+
+    docker:
+      - image: circleci/python:3
```
sciurus: There's an image with the aws cli already you could use if you like.

https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-docker.html#cliv2-docker-install

mythmon: thanks!

mythmon:
bors r= @rehandalal

.circleci/config.yml
mythmon: Turns out that amazon/aws-cli is a bit too minimal for our use cases. I got this error:

> tar utility is not present in this image but it is required. Please install it to have workflow workspace capability.

https://app.circleci.com/pipelines/github/mozilla-extensions/normandy-devtools/2487/workflows/298f5515-1685-4e8a-82ca-2c15a96e927f/jobs/7402

PR #596, opened 2020-11-05
added condition to filter only enrollmentPause false recipes

### LINKED ISSUE #595: Overview Pause Date Recipes do not show correct recipes
Overview Pause Date recipes is filtered from live experiments that have enrollment pause dates, it doesn't filter if the recipe is actually paused

tiftran:
fixes #595


mythmon (approved):
I think Codecov is just being grumpy. This looks good, thanks for noticing the issue.

bors r+

PR #602, opened 2020-11-06
Make UI consistent and decluttering for details page

rehandalal:
this PR does a few things:
- corrects some visual inconsistencies on the details page
  - add subtle text color to field values in approval request and experimenter section
  - fix some margins
- re-orders experimenter details section and approval request section
- declutters the recipe details page header by moving the experimenter and telemetry links
  - links are now in the menu under the ellipsis button
  - links also in the header of experimenter details section
  - links also in the panel of the experimenter details section
- converts a few more `.js` files to `.tsx`

r?

PR #604, opened 2020-11-09
 UI tweaks for overview and recipe listing page 

### LINKED ISSUE #591: Search box on listing page has some unexpected whitespace
image

There's a little unexpected whitespace between the icon on the left and the actual input field.

rehandalal:
Depends on #602 

Fixes #591 

There are a number of things going on here and perhaps room for further refactoring to make things even DRY-er. 

r?


content/components/overview/EndingRecipes.tsx
```diff
@@ -63,50 +56,29 @@ const EndingRecipeCard: React.FC<{
...
-      bordered
-      className="recipe-listing mb-2"
-      header={<CardHeader recipe={recipe} />}
-    >
-      <div className="d-flex flex-wrap m-0">
-        <div className="flex-grow-1">
-          <p>Experimenter End Date: </p>
-          {endingDate.toLocaleDateString()}
+    <RecipeCard recipe={recipe}>
+      <div className="flex-grow-1" />
```
mythmon, line 62: Is this the part that lets it fill up smaller cards to align to the grid? Probably worth commenting it.

content/components/pages/RecipeListingPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+import RecipeQueryEditor from "devtools/components/recipes/listings/RecipeQueryEditor";
+import WriteRecipeModal from "devtools/components/recipes/listings/WriteRecipeModal";
+import {
+  useSelectedEnvironmentState,
+  useSelectedNormandyEnvironmentAPI,
+} from "devtools/contexts/environment";
+import { RecipeListQuery } from "devtools/types/normandyApi";
+import { convertToV1Recipe } from "devtools/utils/recipes";
+
+// Wrapped in a HOC and then exported as default
```
mythmon: This comment isn't accurate anymore.

content/components/recipes/RecipeCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children,
+  className,
+  recipe,
+  ...props
+}) => {
+  const environmentState = useEnvironmentState();
+  const linkUrl = `/${environmentState.selectedKey}/recipes/${recipe.id}/`;
+
+  return (
+    <Panel bordered className={`recipe-listing ${className}`} {...props}>
```
mythmon, line 24: Since classname is often undefined, this ends up producing 

```<div class="recipe-listing undefined rs-panel rs-panel-default rs-panel-bordered">```.

`className` should either have a default of `""`, or we should be using something like `cx` here.

File: content/components/recipes/listings/ListingsHeader.tsx [hunk 4]
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import { Icon, IconButton } from "rsuite";
+
+import { useSelectedEnvironmentState } from "devtools/contexts/environment";
+
+interface ListingsHeaderProps {
+  setShowWriteRecipes: React.Dispatch<React.SetStateAction<boolean>>;
```
mythmon: I think this type is probably too specific. I think all you care about is  `setShowWriteRecipes: (show: boolean) => void`.

content/components/recipes/listings/RecipeList.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { has } from "devtools/utils/helpers";
+import { convertToV1Recipe } from "devtools/utils/recipes";
+
+interface RecipeListProps {
+  recipes: Array<RecipeV3>;
+  copyRecipeToArbitrary: (recipe: RecipeV3) => void;
+  environmentKey: string;
+}
+
+interface CardProps {
```
mythmon: Nit, I think she should be moved to after RecipeList, so it is right next to `RecipeCard`. We don't have a strong pattern for that yet though.

mythmon (commented):
I'm still not really into the big buttons at the bottom. I think the thing that bugs me most is that "Edit" and "Customize and Run" aren't equally common, and this new UI treats them that way. It also doesn't leave much room if we want to add more actions to these cards.

mythmon:
Thanks for the updates. This looks good to me now.

bors r+

PR #611, opened 2020-11-12
Add extensions listing page and convert APIs to TS

mythmon:
Ref #535

PR #628, opened 2020-11-20
Add extensions creation modal

### LINKED ISSUE #535: Add UI for managing extensions
Our earlier plan was to phase out hosted XPIs on Normandy however that is no longer the case and we should build out proper UI for managing them.

mythmon:
Fixes #535
Depends on #611 

I still need to add tests, and I'm having some trouble with styling not showing up in the extension that I was hoping @rehandalal could look at. I added a LESS Module with styles for the extension dialog, and they work fine on the web, but the CSS doesn't seem to get applied at all in the extension.


mythmon:
Here is what it is supposed to look like, and it works fine in the web version of the extension. Notice that the filename is truncated and the box didn't expand off the screen:

image

This is what happens with the extension version. The popup is variable sized, and overflows the screen when long file names are included. There are some other issues too, but that's the major one. It seems like the CSS file just isn't being included at all.

Screenshot from 2020-11-30 14-24-17

mythmon:
With #535 merged, this is ready for review and merging now.

rehandalal (changes requested):
this looks good but needs some conflict resolution due to changes I recently landed.

mythmon:
I've merged in main, this should be ready to merge now.

rehandalal (approved):
bors r+

PR #646, opened 2020-12-09
Add docker-compose file for local Normandy setup

mythmon:
It may make sense to expand this to include Experimenter and potentially other services later, but for now this is already a helpful step for me.


docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+    networks:
+      - public_nw
+      - private_nw
+    environment:
+      DJANGO_DEBUG: "true"
+      DATABASE_URL: "postgres://postgres:postgres@db/normandy"
+      DJANGO_ALLOWED_HOSTS: "*"
+      DJANGO_CONFIGURATION: "ProductionInsecure"
+      DJANGO_SILENCED_SYSTEM_CHECKS: "security.W001,security.W003,security.W004,security.W008,security.W009,security.W012,security.W016,normandy.recipes.E006,dockerflow.health.W001,normandy.recipes.I001,normandy.recipes.I003"
+      DJANGO_AUTOGRAPH_URL: "http://autograph:8001"
```
tiftran: I think maybe there needs to be a slash here? and the port being 8000 too?

I'm getting 
```  File "/opt/venv/lib/python3.9/site-packages/requests/models.py", line 314, in prepare
    self.prepare_url(url, params)
  File "/opt/venv/lib/python3.9/site-packages/requests/models.py", line 382, in prepare_url
    raise InvalidURL(*e.args)
requests.exceptions.InvalidURL: Failed to parse: http://autograph:8001sign/data```

mythmon: Ah right, good catch. What did you do to make that happen? Or what was the rest of the stack trace?

tiftran: Just did a plain old `docker-compose up`

tiftran (approved):

mythmon:
bors r= @tiftran

docker-compose.yml
mythmon: Weird. I was only able to get this to happen by forcing a signing job with manage.py. Oh well, thanks for testing!

PR #647, opened 2020-12-09
Improved UI for Namespace filter

mythmon:
This will make it a lot more reasonable to make this filter a common part of our workflow.


rehandalal (approved):
bors r+

PR #653, opened 2020-12-10
extensionApi details for details page

### LINKED ISSUE #536: Provide more details about extension in addon studies
We currently show the attached id of the extension from the db and no other information about the extension. We should probably expose more details about the extension like the file name, ID, version and maybe a link to the XPI.

tiftran:
Screen Shot 2020-12-10 at 9 05 51 AM

fixes #536 
depends #611

### DIFF
--- a/content/tests/components/recipes/details/RecipeDetails.test.js
+++ b/content/tests/components/recipes/details/RecipeDetails.test.js
@@ -188,11 +188,12 @@ describe("The `RecipeDetails` component", () => {
     }
 
     for (const extension of extensions) {
-      const nameRegex = new RegExp(extension.name);
-      const versionRegex = new RegExp(extension.version);
-
-      expect(doc.getByText(nameRegex)).toBeInTheDocument();
-      expect(doc.getByText(versionRegex)).toBeInTheDocument();
+      expect(
+        doc.getByText(extension.name, { exact: false }),
+      ).toBeInTheDocument();
+      expect(
+        doc.getByText(extension.version, { exact: false }),
+      ).toBeInTheDocument();
     }
   });
 
--- a/content/tests/components/recipes/details/RecipeDetails.test.js
+++ b/content/tests/components/recipes/details/RecipeDetails.test.js
@@ -6,6 +6,7 @@ import RecipeDetailsPage from "devtools/components/pages/RecipeDetailsPage";
 import RecipeDetails from "devtools/components/recipes/details/RecipeDetails";
 import { RecipeDetailsProvider } from "devtools/contexts/recipeDetails";
 import { actionFactory } from "devtools/tests/factories/api";
+import { extensionFactory } from "devtools/tests/factories/api";
 import { experimenterResponseFactory } from "devtools/tests/factories/experiments";
 import {
   versionFoFactory,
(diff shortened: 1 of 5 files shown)


content/tests/components/recipes/details/RecipeDetails.test.js
```diff
@@ -94,6 +95,23 @@ describe("The `RecipeDetails` component", () => {
     return /** @type import("devtools/types/recipes").RecipeV3<import("devtools/types/arguments").BranchedAddonStudyArguments> */ (recipe);
   };
 
+  const extensionSetup = () => {
+    const extensions = extensionFactory.buildCount(2);
+    const mockFetchExtension = jest
+      .fn()
+      .mockImplementation((i) => extensions[i % 3]);
```
mythmon: `i % 3` here probably isn't right, since there are only two extensions. This should probably be `.mockImplementation((i) => extensions[i % extensions.length]);`

content/tests/components/recipes/details/RecipeDetails.test.js
```diff
@@ -166,6 +186,14 @@ describe("The `RecipeDetails` component", () => {
       ).not.toHaveLength(0);
       expect(doc.getAllByText(branch.ratio.toString())).not.toHaveLength(0);
     }
+
+    for (const extension of extensions) {
+      const nameRegex = new RegExp(extension.name);
+      const versionRegex = new RegExp(extension.version);
+
+      expect(doc.getByText(nameRegex)).toBeInTheDocument();
+      expect(doc.getByText(versionRegex)).toBeInTheDocument();
```
mythmon: Why use regex here, instead of just using the name and version directly in the `getByText` call?

mythmon (changes requested):

content/tests/components/recipes/details/RecipeDetails.test.js
tiftran: ooh right right

content/tests/components/recipes/details/RecipeDetails.test.js
tiftran: i mushed the name and version together so that it's `<name> (version : <version>)` and `getByText` needs a regex for substrings?

mythmon: Ah. That will work, but there is a simpler way: `expect(doc.getByText(extension.name, {exact: false})`. That will let it match partial elements.

mythmon:
A while ago Rehan found a problem where some code patterns get broken by coverage mode somehow. I was able to get the test failure CI is seeing with `yarn test --silent --coverage`.

content/components/recipes/form/arguments/MultiPreferenceBranches.js
```diff
@@ -41,8 +41,8 @@ export default function MultiPreferenceBranches() {
   };
 
   let branchesList = <HelpBlock>There are no branches.</HelpBlock>;
-  const branchLength = data.arguments.branches.length;
   if (data.arguments.branches && data.arguments.branches.length) {
```
tiftran: not sure why I originally had it this way... 😬

mythmon:
bors d+

This is good to go, it just has some conflicts to resolve now.

PR #654, opened 2020-12-11
Make some acronyms render better in automatic labels

mythmon:
This is entirely cosmetic, but it's something that has bugged me for a while. This fixes the cases I know about, but maybe there are others we should add to the list?


leplatrem (approved):

mythmon:
bors r= @leplatrem

PR #672, opened 2021-02-01
Make console.warn and console.error throw an error for new tests

### LINKED ISSUE #534: Test output is extremely noisy

rehandalal:
Fix #534.

This is the stopgap solution we discussed to ensure new tests don't make these `console.warn` and `console.error` calls. It will allow us to slowly fix the existing tests and hopefully eventually totally clean up the output.

r?

### DIFF
--- a/content/tests/.eslintrc.js
+++ b/content/tests/.eslintrc.js
@@ -1,6 +1,8 @@
 module.exports = {
   globals: {
     renderWithContext: false,
+    restoreConsole: false,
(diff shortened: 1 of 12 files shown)


mythmon (approved):
bors r+

PR #674, opened 2021-02-05
Add history timeline sidebar

### LINKED ISSUE #252: Add recipe history timeline
We need UI to expose the history of recipes and allow us to navigate through previous revisions. This should probably take the form of a timeline of revisions with nested nodes that outline the enable state history. It should also expose the revision comments and highlight the currently viewed revision and the current live revision.

rehandalal:
Fixes #252 

r?


mythmon (changes requested):
I'd like to see some tests for the new behavior. Test coverage wasn't hurt, but none of the tests verify any of the new code, they just run it.

Besides that, this looks like a good base that gets the vital information in. We can build on it from here to add nice-to-haves.

rehandalal:
I filed an issue #677 to re-add tests that I removed

mythmon (approved):

PR #682, opened 2021-02-06
Add more details to revision tooltip

### LINKED ISSUE #675: Add more details to history timeline revision tooltips
The new history timeline has a little tooltip for each revision that currently shows the comment for the revision. We should update these to have more details like the time the revision was created, who made the revision and the comment.

rehandalal:
Fixes #675 

r?


mythmon (approved):
lgtm, though there is a test failure that will need fixed.

bors d+

mythmon:
I think the test that was failing just takes a long time. It passes for me locally, but it is the longest test of the group. Artificially lowering my timeout from the default 5 seconds to one seconds with `--testTimeout 1000` made the test fail.

I'm looking into how we can make this test faster, or possibly to break it up, but for now I think we should just give it some extra time.

bors r+

PR #685, opened 2021-02-08
Memoize Normandy and Experimenter API hooks

### LINKED ISSUE #683: Action selector re-requests actions on every key press in create recipe page
On the recipe creation page, typing in any of the text boxes (like the extra filter expression), causes the action selector to re-fetch the list of actions from the server on every key press. That's bad.

image

mythmon:
Fixes #683

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@babel/preset-typescript": "7.12.7",
     "@testing-library/jest-dom": "5.11.6",
     "@testing-library/react": "11.2.2",
-    "@testing-library/react-hooks": "^5.0.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -41,6 +41,7 @@
     "@babel/preset-typescript": "7.12.7",
     "@testing-library/jest-dom": "5.11.6",
     "@testing-library/react": "11.2.2",
+    "@testing-library/react-hooks": "^5.0.3",
```
rehandalal: sorry!!!

nit: pin

rehandalal (approved):
bors d=@mythmon

PR #757, opened 2022-02-01
Update to Yarn v3

tiziodcaio:
Hi I upgraded yarn to version 3. I migrated what is necessary and I hop it will works.
[Here why we should upgrade](https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern)
I used [corepack](https://nodejs.org/dist/latest-v16.x/docs/api/corepack.html) to do the upgrade, so you have not to bundle always the yarn js script file. The yarn caches are saved by circleci instead of saving directly the node_modules folder.


tiftran:
Hi @tiziodcaio!

Normandy devtools is currently in maintenance mode so we aren’t actively doing any development work on it other than potential critical bug fixes. We appreciate your effort in contributing to Mozilla’s open-source projects and we encourage you to look for other projects that might benefit from your contributions.

